PR #2, opened 2016-04-12
JavaScriptContext

RedEnchilada:
Adds a context for assertions related to JavaScript variables.


src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -0,0 +1,54 @@
...
+     * @Then the javascript variable :variable should not be null
+     */
+    public function assertJavascriptVariableHasAValue($variable)
+    {
+        // Get the value of our variable from javascript
+        $result = $this->getSession()->evaluateScript('return ' . $variable . ';');
+
+        // If it's empty - we failed
+        if (empty($result)) {
+            throw new ExpectationException('The custom variable "' . $variable . '" is null or does not exist.', $this->getSession());
```
crschoen: This line exceeds the 120 character limit. Please add line breaks where appropriate.

RedEnchilada:
Addressed @crschoen

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -0,0 +1,54 @@
...
+        // Get the type of our variable from javascript.
+        $result = $this->getSession()->evaluateScript('return typeof(' . $variable . ');');
+
+        // If it doesn't match - we failed.
+        if ($result != $type xor $not) {
+            $notnot = $not ? '' : ' not';
+            throw new ExpectationException(
+                "The variable \"$variable\" should$not be type $type, but was$notnot",
+                $this->getSession()
+            );
```
crschoen: This could be cleaned up by eliminating double `$not` as follows:
```php
if ($result != $type xor $not) {
    throw new ExpectationException(
        "The variable type of \"$variable\" should$not be type $type, but it is $result",
        $this->getSession()
    );
}
```
This also has the benefit of communicating the actual type of the variable.

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
```diff
@@ -14,4 +16,13 @@
      * @param string $text Text to be searched in the page.
      */
     abstract public function assertPageContainsText($text);
+
+    /**
+     * Returns Mink session.
+     *
+     * @param string|null $name name of the session OR active session will be used
+     *
```
crschoen: Please remove this *blank* line.

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
```diff
@@ -14,4 +16,13 @@
...
+     * @return Session
+     */
+    abstract public function getSession($name = null);
```
crschoen: Please add a `@param` annotation for the `$name` variable.

src/Behat/FlexibleMink/PseudoInterface/JavaScriptContextInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ * Functionality for testing JavaScript variables.
+ */
+trait JavaScriptContextInterface
+{
+    /**
+     * Determines if a javascript variable is set and has a value.
+     *
+     * @throws ExpectationException if the given variable is undefined.
+     */
+    abstract public function assertJavascriptVariableHasAValue($variable);
```
crschoen: Please add a `@param` annotation for the `$variable` variable.

src/Behat/FlexibleMink/PseudoInterface/JavaScriptContextInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Determines if a javascript variable matches a specific type.
```
crschoen: Determine if **the type of** a javascript variable matches a specific type.

src/Behat/FlexibleMink/PseudoInterface/JavaScriptContextInterface.php
```diff
@@ -0,0 +1,27 @@
...
+     *
+     * @param  string               $variable The variable to evaluate type for.
+     * @param  string               $not      Invert the check?
```
crschoen: Please indicate that this is evaluated as a bool. The reader could get confused and think this is a `'yes'` or `'no'` option or perhaps a `'true'` or `'false'` option.

src/Behat/FlexibleMink/PseudoInterface/JavaScriptContextInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Behat\FlexibleMink\PseudoInterface;
+use Behat\Mink\Exception\ExpectationException;
```
crschoen: Please add a blank line between the namespace and the import statements.

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Behat\FlexibleMink\Context;
+
+use Behat\FlexibleMink\PseudoInterface\FlexibleContextInterface;
+use Behat\FlexibleMink\PseudoInterface\JavaScriptContextInterface;
+use Behat\Mink\Exception\ExpectationException;
+
+trait JavaScriptContext
```
crschoen: Please add a docblock for this trait. An `{@inheritdoc}` tag should be enough.

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        // Get the value of our variable from javascript
+        $result = $this->getSession()->evaluateScript('return ' . $variable . ';');
+
+        // If it's empty - we failed
+        if (empty($result)) {
+            throw new ExpectationException('The custom variable "' . $variable . '" is null or does not exist.', $this->getSession());
+        }
+
+        return true;
```
crschoen: Assertions do not need a return. The assertion only needs to complete without throwing an error.

crschoen:
First code review complete. @RedEnchilada 	e09fa27

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
RedEnchilada: There's already one there :cat:

RedEnchilada:
Comments have been addressed @crschoen

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
crschoen: Not sure how I missed that. XD

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -0,0 +1,57 @@
...
+    // Depends.
+    use FlexibleContextInterface;
+
+    // Implements.
+    use JavaScriptContextInterface;
+
+    /**
+     * {@inheritdoc}
+     *
+     * @Then the javascript variable :variable should not be null
```
Chekote: I know this has already been merged, but I just wanted to point out that this sentence does not match up exactly with what the method body does.

The method body is checking whether or not the variable is empty, which is subtly different from null. For example, an empty string or an undefined variable would pass this assertion, even though neither are null.

crschoen: #3 addresses this.

PR #4, opened 2016-04-13
Change the not null test to actually check for not null

### LINKED ISSUE #3: Inaccurate description for assertJavascriptVariableHasAValue
The method body is checking whether or not the variable is empty, which is subtly different from null. For example, an empty string or an undefined variable would pass this assertion, even though neither are null.

RedEnchilada:
instead of using empty (which accomplishes a different purpose)

This branch name may be longer than the diff file.

(The tests that use this step will need tweaked a bit, but that's fine because the tests in question on our current projects are poorly-worded anyway. This will make them clearer (: )

Resolves #3

### DIFF
--- a/src/Behat/FlexibleMink/Context/JavaScriptContext.php
+++ b/src/Behat/FlexibleMink/Context/JavaScriptContext.php
@@ -28,7 +28,7 @@ trait JavaScriptContext
         $result = $this->getSession()->evaluateScript('return ' . $variable . ';');
 
         // If it's null - we failed
-        if ($result == null) {
+        if ($result === null) {
             throw new ExpectationException(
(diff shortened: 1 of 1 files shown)


src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -27,8 +27,8 @@ public function assertJavascriptVariableHasAValue($variable)
         // Get the value of our variable from javascript
         $result = $this->getSession()->evaluateScript('return ' . $variable . ';');
 
-        // If it's empty - we failed
-        if (empty($result)) {
+        // If it's null - we failed
+        if ($result == null) {
```
crschoen: Shouldn't this be a strict comparison? `===`

PR #10, opened 2016-04-20
Added ability to compare stored thing with a value.

Chekote:
This is required for a test being written for stdcheck.com.

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -28,6 +28,19 @@ trait StoreContext
 
     /**
      * {@inheritdoc}
+     *
+     * @Then /^the "(?P<key>[^"]+)" should be "(?P<value>[^"]*)"$/
+     * @Then /^the "(?P<key>[^"]+)" should be empty$/
+     */
+    public function assertThingIs($key, $value = null)
+    {
+        if ($this->get($key) != $value) {
+            throw new Exception("Expected $key to be " . var_export($value, true) . ', but it was not');
+        }
+    }
+
+    /**
+     * {@inheritdoc}
      */
     protected function put($thing, $key)
     {
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -29,13 +29,14 @@ trait StoreContext
     /**
      * {@inheritdoc}
      *
-     * @Then /^the "(?P<key>[^"]+)" should be "(?P<value>[^"]*)"$/
-     * @Then /^the "(?P<key>[^"]+)" should be empty$/
+     * @Then /^the "(?P<key>[^"]+)" should be (?P<value>.*)$/
      */
-    public function assertThingIs($key, $value = null)
+    public function assertThingIs($key, $expected = null)
     {
-        if ($this->get($key) != $value) {
-            throw new Exception("Expected $key to be " . var_export($value, true) . ', but it was not');
+        if (($actual = $this->get($key)) !== $expected) {
+            throw new Exception(
+                "Expected $key to be " . var_export($expected, true) . ', but it was ' . var_export($actual, true)
+            );
         }
     }
 
--- a/src/Behat/FlexibleMink/Context/TypeCaster.php
+++ b/src/Behat/FlexibleMink/Context/TypeCaster.php
@@ -36,13 +36,13 @@ trait TypeCaster
      *
      * Supports true and false only. e.g. will not cast 0 or 1.
      *
-     * @Transform /^(true|false)$/
+     * @Transform /^(true|false)$/i
      * @param  string $string the string to cast.
      * @return bool   The resulting bool.
      */
     public function castStringToBool($string)
     {
-        return $string === 'true';
+        return strtolower($string) === 'true';
     }
 
     /**
--- a/src/Behat/FlexibleMink/Context/TypeCaster.php
+++ b/src/Behat/FlexibleMink/Context/TypeCaster.php
@@ -22,7 +22,7 @@ trait TypeCaster
     /**
      * Casts a step argument from a string to a float.
      *
-     * @Transform /^(\d+)\.(\d+)$/
+     * @Transform /^(\d*)\.(\d+)$/
      * @param  string $string the string to cast.
      * @return float  The resulting float.
      */
--- a/src/Behat/FlexibleMink/Context/TypeCaster.php
+++ b/src/Behat/FlexibleMink/Context/TypeCaster.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Behat\FlexibleMink\Context;
+
+/**
+ * Adds type casting for step arguments.
+ */
+trait TypeCaster
+{
+    /**
+     * Casts a step argument from a string to an int.
+     *
+     * @Transform /^(\d+)$/
+     * @param  string $string the string to cast.
+     * @return int    The resulting int.
+     */
+    public function castStringToInt($string)
+    {
+        return intval($string);
+    }
+
+    /**
+     * Casts a step argument from a string to a float.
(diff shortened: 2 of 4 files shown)


src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -27,6 +28,19 @@ public function clearRegistry()
...
     /**
      * {@inheritdoc}
+     *
+     * @Then /^the "(?P<key>[^"]+)" should be "(?P<value>[^"]*)"$/
+     * @Then /^the "(?P<key>[^"]+)" should be empty$/
+     */
+    public function assertThingIs($key, $value = null)
+    {
+        if ($this->get($key) != $value) {
+            throw new Exception("Expected $key to be " . var_export($value, true) . ', but it was not');
```
crschoen: An `ExpectationException` would be more specific.

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -27,6 +28,19 @@ public function clearRegistry()
 
     /**
      * {@inheritdoc}
+     *
+     * @Then /^the "(?P<key>[^"]+)" should be "(?P<value>[^"]*)"$/
+     * @Then /^the "(?P<key>[^"]+)" should be empty$/
+     */
+    public function assertThingIs($key, $value = null)
+    {
+        if ($this->get($key) != $value) {
```
crschoen: I think this should be a strict check `!==` since the values could be something falsy, like `'0'`, `null` or `false`, but not the same.

crschoen:
First code review complete. @Chekote 06a547f

src/Behat/FlexibleMink/Context/StoreContext.php
Chekote: I chose not to use ExpectationException because that class is tightly coupled with MinkContext, which means that you would not be able to use this StoreContext trait in any other Context. We already do this with the APIContext in stdcheck, so making that change would make FlexibleMink incompatible with stdcheck.

src/Behat/FlexibleMink/Context/StoreContext.php
Chekote: I chose not to do a strict check because the data is often coming in from a feature file, which does not offer much in the way of type casting.

I can certainly do that, but I would need to implement additional regexes to match inputs such as string, int, bool, float etc. Since all data coming from feature files are strings, even if there are no quotes around the arguments, I would then have to implement [step argument transformations](http://docs.behat.org/en/v3.0/guides/2.definitions.html#step-argument-transformations) to type cast the incoming values.

Do you think that this should be done as part of this pull request, or could it be a feature request / spike for this project?

Chekote:
@crschoen I have responded to all of the code review comments. Please let me know your thoughts. Thanks.

src/Behat/FlexibleMink/Context/StoreContext.php
crschoen: The only requirement that I see in the `ExpectationException` class is that it has a `Session` or `Driver` instance to work with. This comes from the `Mink` instance. I see now that the `APIContext` uses phpunit. I guess if we're going to write modules that can't use mink exceptions we probably shouldn't call this a mink extension, but more generically a behat extension.

src/Behat/FlexibleMink/Context/StoreContext.php
crschoen: Why not cast the value retrieved from the store as a string before doing the comparison? I'm thinking specifically of the case where there is a stored value that is `0`. If a loose comparison is used the `...should be empty` step would still pass even though there shouldn't be a value.

src/Behat/FlexibleMink/Context/StoreContext.php
Chekote: Yes, that makes sense.

src/Behat/FlexibleMink/Context/StoreContext.php
Chekote: Because I wanted to fully support loose comparison to allow for developers to call the method directly instead of going through a feature file. The current implementation supports loose comparison with strings, numbers and bools via feature files, and additionally supports loose comparison of arrays and objects via direct invocation.

crschoen: I thought that might be what was going on. You could have two methods, one with the step definition that is used for string comparison only and then the method that actually does the comparison which is to be used for internal calls. It's up to you on how you want to proceed. Since this is the only pending comment, merging of this branch is at your discretion.

src/Behat/FlexibleMink/Context/StoreContext.php
crschoen: Yeah, we'll just merge this and refactor into a behat extension later.

crschoen:
👍

src/Behat/FlexibleMink/Context/StoreContext.php
Chekote: I see what you mean. To address your other concern of writing less code, I would prefer to keep it to a single method and instead implement the step argument transformations along with your recommendation of using strict comparison so we can have less code *and* support both requirements.

Chekote:
I added a commit to implement strict comparison using [step argument transformations](http://docs.behat.org/en/v3.0/guides/2.definitions.html#step-argument-transformations), please let me know what you think. Thanks!

src/Behat/FlexibleMink/Context/TypeCaster.php
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * Supports true and false only. e.g. will not cast 0 or 1.
+     *
+     * @Transform /^(true|false)$/
+     * @param  string $string the string to cast.
+     * @return bool   The resulting bool.
+     */
+    public function castStringToBool($string)
+    {
+        return $string === 'true';
```
crschoen: What do you think about changing this to a case insensitive match and adjusting the pattern accordingly?

src/Behat/FlexibleMink/Context/TypeCaster.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public function castStringToInt($string)
+    {
+        return intval($string);
+    }
+
+    /**
+     * Casts a step argument from a string to a float.
+     *
+     * @Transform /^(\d+)\.(\d+)$/
```
crschoen: What about floats that omit the leading zeroes? `.05` etc.
```php
* @Transform /^(\d*)\.(\d+)$/
```

crschoen:
I like the transformations, but I think it will make documentation for argument data types a challenge. Since the argument data type can change depending on the pattern I'm not sure how to document them in the pseudo interfaces. `mixed` maybe?

crschoen:
Second code review complete. @Chekote 287669c

src/Behat/FlexibleMink/Context/TypeCaster.php
Chekote: Good catch.

src/Behat/FlexibleMink/Context/TypeCaster.php
Chekote: I'm not completely sold on that as I couldn't think of a scenario where you would write a sentence with the word true or false with an uppercase letter. But I implemented it anyway so we at least have the option.

Chekote:
In regards to the argument documentation, in preparation for stronger type enforcement via PHP 7, I think they should be documented to reflect the datatype that *should* come into the function, not what *can*. When we move to PHP 7 we can then strictly *require* these types, and allow the language to error out if someone were to write a test (either via fixture file or a direct function call) that passes in the wrong data type.

Chekote:
All comments have been addressed. Ready for next review. bf3f666

Thanks for the feedback on this. This is a much better addition now.

crschoen:
Looks great! 👍

PR #12, opened 2016-04-22
Added Step Definition for Storing a new Reference

crschoen:
This new step definition allows a stored thing to be referred to by a name called out in the scenario. For instance, if there is a step that stores a user under the generic name "user" (`Given there is a user`), that user could then be referred to by name (`And I refer to the "user" as "John"`). This might be done when wanting to create multiple instances of a user each with different names.


awesomeousPrime:
👍

PR #19, opened 2016-05-06
CSV/download checking utilities

RedEnchilada:
I stole these contexts from Donald's pull request elsewhere to put them there lol

Some new contexts for checking the contexts of a CSV download. I opted not to `use` them by default and let the user import them if they want the functionality, since they aren't 100% related to the base FlexibleContext's goalset, but if we'd rather import them by default it'd be two extra lines to add that.


awesomeousPrime:
👍

PR #21, opened 2016-05-10
changing assertIsStored to return store object instead of bool

awesomeousPrime:
assertIsStored is expected to return an object but at the moment is returning a boolean value.


RedEnchilada:
👍

PR #22, opened 2016-05-23
Add function to assert if option exists in the select

johnnie502:
Usage:

`the` _optionName_ `option` [_not_] `exist`[s] `in the` _selectName_ `select`

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -163,7 +163,7 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      *
-     * @Then /^the (?P<option>.*?) option(?P<existence>( | not)?) exists? in the (?P<select>.*?) select$/
+     * @Then /^the (?P<option>.*?) option(?:|(?P<existence> does not?)) exists? in the (?P<select>.*?) select$/
      */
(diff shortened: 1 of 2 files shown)


johnnie502:
!assign back-end @Chekote

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -159,4 +159,23 @@ public function assertPageNotContainsFields(TableNode $tableNode)
             $this->assertFieldNotExists($field);
         }
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @Then /^the (?P<option>.*?) option(?P<existence>( | not)?) exists? in the (?P<select>.*?) select$/
```
RedEnchilada: `(?P<existence>( | not)?)` is kind of a strange regex; it attempts to match a space, which would cause a check for existence to need double spacing, and the ? makes it optional, which doesn't line up with the normal syntax we use for regular expressions. (Also, "the option not exists" would be better worded as "the option does not exist".)

A better (more consistent with our naming schemes) version of this particular block would be something like `(?:|(?P<existence> does not))`

RedEnchilada:
Code review complete 96e5edb

johnnie502:
@RedEnchilada ready for 2nd review

RedEnchilada:
👍

PR #24, opened 2016-05-24
Allow behat to modify cookies through Mink.

bluefalcon26:
Adds the functions
`addOrReplaceCookie($key, $value)`
and
`deleteCookie($key)`
to FlexibleMink class.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -181,8 +181,9 @@ class FlexibleContext extends MinkContext
 
     /**
      * Adds or replaces a cookie.
+     * Note that you must request a page before trying to set a cookie, in order to set the domain.
      *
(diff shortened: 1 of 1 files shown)


src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -178,4 +178,26 @@ public function assertSelectContainsOption($select, $existence, $option)
             throw new ExpectationException("The option '" . $option . "' not exist in the select", $this->getSession());
         }
     }
+
+    /**
+     * Adds or replaces a cookie.
+     *
+     * @When /^(?:|I )set the cookie "(?P<key>(?:[^"]|\\")*)" with value "(?P<value>(?:[^"]|\\")*)"$/
```
RedEnchilada: For `value`, it might be better to remove the quotes surrounding the captured block and let it capture anything, so that the step argument transformations elsewhere in the project can cast the value as an int, bool, or whatever else it handles. So it'd be a simple `with value (?P<value>.+)`.

RedEnchilada:
Code review complete. Nice additions, particularly for other step definitions to use! 👍

RedEnchilada:
👍

PR #26, opened 2016-05-26
add function to ensure angular is loaded and page is rendered

johnnie502:
This will ensure the angular is loaded and the page is rendered when testing angular based project.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -201,4 +201,22 @@ class FlexibleContext extends MinkContext
         // set cookie:
         $this->getSession()->setCookie($key, null);
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @When angular rendered the page
+     */
+    public function waitForAngular()
+    {
+        // Wait for angular to load
+        $this->getSession()->wait(5000, "typeof angular != 'undefined'");
+        // Wait for angular to be testable
+        $this->getSession()->evaluateScript(
+            'angular.getTestability(document.body).whenStable(function() {
+                window.__testable = true;
+            })'
+        );
+        $this->getSession()->wait(5000, 'window.__testable == true');
+    }
 }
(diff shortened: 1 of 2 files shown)


src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -201,4 +201,22 @@ public function deleteCookie($key)
         // set cookie:
         $this->getSession()->setCookie($key, null);
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @When angular rendered the page
+     */
+    public function waitForAngular()
```
Chekote: I'm not sure this is appropriate in FlexibleMink. FlexibleMink is an extension of Mink, which itself is a general purpose tool for working with browsers. It shouldn't have any javascript specific functionality. This should really be in a separate composer package specifically for working with Angular.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -201,4 +201,22 @@ public function deleteCookie($key)
...
+    {
+        // Wait for angular to load
+        $this->getSession()->wait(5000, "typeof angular != 'undefined'");
```
Chekote: I would imagine that you'd need an error trap here for if the timeout passed and angular was still not defined. Otherwise you'll get a JS error when you call angular on line 216.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -201,4 +201,22 @@ public function deleteCookie($key)
...
+        // Wait for angular to be testable
+        $this->getSession()->evaluateScript(
+            'angular.getTestability(document.body).whenStable(function() {
+                window.__testable = true;
+            })'
+        );
+        $this->getSession()->wait(5000, 'window.__testable == true');
```
Chekote: Again, I think you'll need an error trap here for if this timeout passes and __testable is still not true.

Chekote:
Code review for aadcfca is complete.

Issue #27, opened 2016-05-26
Add support for testing JavaScript JSON objects key-by-key


hopeseekr:
**Problem:** Testing for exact JSON objects is fraught with many grave difficulties, such as orderIds, dates and other dynamic data constantly changing.

**Need:** The ability to compare two JSON objects, one from JavaScript, one from the Behat tests, on a key-by-key basis, including the keys one wants to test, and ignoring the rest.

RedEnchilada:
Resolved by #28

PR #28, opened 2016-05-26
(#27) Added support for testing JS-sourced JSON objects key-by-key.

hopeseekr:
**Problem:** Testing for exact JSON objects is fraught with many grave difficulties, such as orderIds, dates and other dynamic data constantly changing.

**Solution:** This adds the ability to compare two JSON objects, one from JavaScript, one from the Behat tests, on a key-by-key basis, including the keys one wants to test, and ignoring the rest.

### DIFF
--- a/src/Behat/FlexibleMink/Context/JavaScriptContext.php
+++ b/src/Behat/FlexibleMink/Context/JavaScriptContext.php
@@ -59,7 +59,7 @@ trait JavaScriptContext
     /**
      * {@inheritdoc}
      *
-     * @Given /^the javascript variable :variableName should have the following contents:$/
+     * @Given the javascript variable :variableName should have the following contents:
      */
     public function assertJsonContentsOneByOne($variableName, TableNode $values)
     {
--- a/src/Behat/FlexibleMink/Context/JavaScriptContext.php
+++ b/src/Behat/FlexibleMink/Context/JavaScriptContext.php
@@ -57,19 +57,22 @@ trait JavaScriptContext
     }
 
     /**
-     * Selectively compares two JSON objects.
+     * {@inheritdoc}
      *
-     * @Given /^the javascript variable "([^"]*)" should have the following contents:$/
+     * @Given /^the javascript variable :variableName should have the following contents:$/
      */
-    public function assertJsonContentsOneByOne($variableName, TableNode $values, $count = null)
+    public function assertJsonContentsOneByOne($variableName, TableNode $values)
     {
-        $returnedJsonData = $this->getSession()->evaluateScript('return JSON.stringify(' . $variableName . ');');
+        $returnedJsonData = $this->getSession()->evaluateScript(
+            'return JSON.stringify(' . $variableName . ');'
+        );
         $response = json_decode($returnedJsonData, true);
 
         foreach ($values->getHash() as $row) {
             if (!isset($response[$row['key']])) {
                 throw new ExpectationException(
-                    sprintf('Expected key "%s" was not in the JS variable "%s"\nActual: %s', $row['key'], $variableName, $returnedJsonData),
+                    "Expected key \"{$row['key']}\" was not in the JS variable \"{$variableName}\"\n" .
+                        "Actual: $returnedJsonData",
                     $this->getSession()
                 );
             }
@@ -78,13 +81,20 @@ trait JavaScriptContext
 
             if ($actual != $expected) {
                 throw new ExpectationException(
-                    sprintf('Expected "%s" in %s position but got "%s"', $expected, $row['key'], $actual),
+                    "Expected \"$expected\" in {$row['key']} position but got \"$actual\"",
                     $this->getSession()
                 );
             }
         }
     }
 
+    /**
+     * Returns as-is literal inputs (string, int, float), otherwise
+     * returns the JSON encoded output.
+     *
+     * @param  mixed  $value
+     * @return string JSON encoded string
+     */
     protected function getRawOrJson($value)
     {
         if (is_array($value) || is_object($value)) {
--- a/src/Behat/FlexibleMink/Context/JavaScriptContext.php
+++ b/src/Behat/FlexibleMink/Context/JavaScriptContext.php
@@ -4,6 +4,7 @@ namespace Behat\FlexibleMink\Context;
 
 use Behat\FlexibleMink\PseudoInterface\FlexibleContextInterface;
 use Behat\FlexibleMink\PseudoInterface\JavaScriptContextInterface;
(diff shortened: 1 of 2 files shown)


src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
             );
         }
     }
+
+    /**
+     * Selectively compares two JSON objects.
+     *
+     * @Given /^the javascript variable "([^"]*)" should have the following contents:$/
+     */
+    public function assertJsonContentsOneByOne($variableName, TableNode $values, $count = null)
```
Chekote: Please add `@param` annotations for the parameters. When you do so, please also describe the required columns for the TableNode. Thanks!

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
             );
         }
     }
+
+    /**
+     * Selectively compares two JSON objects.
+     *
+     * @Given /^the javascript variable "([^"]*)" should have the following contents:$/
```
Chekote: Since this pattern only requires a simple parameter (a quoted string), you can use the simple declaration syntax:

`@Given the javascript variable :variableName should have the following contents:`

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
...
+            if ($actual != $expected) {
+                throw new ExpectationException(
+                    sprintf('Expected "%s" in %s position but got "%s"', $expected, $row['key'], $actual),
+                    $this->getSession()
+                );
+            }
+        }
+    }
+
+    protected function getRawOrJson($value)
```
Chekote: Please add a docblock for this method describing it's purpose, and documenting it's parameters and return type/value. Thanks!

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
...
+     */
+    public function assertJsonContentsOneByOne($variableName, TableNode $values, $count = null)
+    {
+        $returnedJsonData = $this->getSession()->evaluateScript('return JSON.stringify(' . $variableName . ');');
+        $response = json_decode($returnedJsonData, true);
+
+        foreach ($values->getHash() as $row) {
+            if (!isset($response[$row['key']])) {
+                throw new ExpectationException(
+                    sprintf('Expected key "%s" was not in the JS variable "%s"\nActual: %s', $row['key'], $variableName, $returnedJsonData),
```
Chekote: This line is a little too long. We want to ideally keep the right margin at 120 chars or less.

Since you're doing simple parameter injection in these strings, an echo would be more readable here. Using sprintf is going to increase the difficulty of reading this code because the developer is going to have to determine which %s each parameter maps to. I think the following would be more readable:

```php
throw new ExpectationException(
    "Expected key '{$row['key']}' was not in the JS variable '$variableName'\n" .
        "Actual: $returnedJsonData',
    $this->getSession()
);
```

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
...
     }
+
+    /**
+     * Selectively compares two JSON objects.
+     *
+     * @Given /^the javascript variable "([^"]*)" should have the following contents:$/
+     */
+    public function assertJsonContentsOneByOne($variableName, TableNode $values, $count = null)
+    {
+        $returnedJsonData = $this->getSession()->evaluateScript('return JSON.stringify(' . $variableName . ');');
```
Chekote: This line is a little too long. We want to ideally keep the right margin at 120 chars or less. I suggest the following:

```php
$returnedJsonData = $this->getSession()->evaluateScript(
	'return JSON.stringify(' . $variableName . ');'
);
```

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
...
+                    sprintf('Expected key "%s" was not in the JS variable "%s"\nActual: %s', $row['key'], $variableName, $returnedJsonData),
+                    $this->getSession()
+                );
+            }
+            $expected = $this->getRawOrJson($row['value']);
+            $actual = $this->getRawOrJson($response[$row['key']]);
+
+            if ($actual != $expected) {
+                throw new ExpectationException(
+                    sprintf('Expected "%s" in %s position but got "%s"', $expected, $row['key'], $actual),
```
Chekote: I suggest making the same changes for this line that I mentioned [here](https://github.com/Medology/FlexibleMink/pull/28/files#r64905451).

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -54,4 +55,42 @@ public function assertJavascriptVariableType($variable, $not, $type)
...
+                    $this->getSession()
+                );
+            }
+        }
+    }
+
+    protected function getRawOrJson($value)
+    {
+        if (is_array($value) || is_object($value)) {
```
Chekote: Why do we need to make this distinction? Why not json_encode everything?

Chekote:
Code review for 65cae94 is complete.

src/Behat/FlexibleMink/Context/JavaScriptContext.php
hopeseekr: Because in the test case, you can put `asdf` instead of `"asdf"`.

hopeseekr:
Ready for code review again.

src/Behat/FlexibleMink/Context/JavaScriptContext.php
Chekote: It looks like you left the `/^` and `$/` in the pattern. I don't believe that the pattern will match with that. When writing these, you can either use a full on regex (as you did originally), or you can use the Cucumber syntax that only supports :varname.

hopeseekr:
Ready for code review.

PR #32, opened 2016-06-07
Fix scrollBodyTo won't work with floating elements

Chekote:
__Method__
Create a HTML document with a floating element that extends beyond the bottom of the page.
Call scrollBodyTo('bottom').

__Expected Behavior__
The window should scroll to the bottom of the page.

__Actual Behavior__
The window does not scroll at all.

__Cause__
The floating element breaks the javascript scrollIntoView function.

__Resolution__
I changed the method to use window.scrollTo instead, which works fine with floating elements.


Chekote:
Just an FYI: This is another way to accomplish scrolling a specific element to the top or bottom of the viewport:

```js
element.scrollTop = element.scrollHeight - element.clientHeight
```

So for body, you can do:

```js
document.body.scrollTop = document.body.scrollHeight - document.body.clientHeight
```

crschoen:
👍

PR #34, opened 2016-06-07
making change to link downloads

awesomeousPrime:
Changing downloadViaLink to also include an optional headers argument.


johnnie502:
👍  good to merge

PR #35, opened 2016-06-07
prepend base_url if needed

awesomeousPrime:
Adding base_url if one is not included in url for visible links.

### DIFF
--- a/src/Behat/FlexibleMink/Context/WebDownloadContext.php
+++ b/src/Behat/FlexibleMink/Context/WebDownloadContext.php
@@ -18,31 +18,33 @@ trait WebDownloadContext
     use FlexibleContextInterface;
     use StoreContextInterface;
 
+    protected static $baseUrlRegExp = '/^((http(s|):[\/]{2}|)([a-zA-Z]+\.|)[a-zA-Z0-9]+\.[a-zA-Z]+(\:[\d]+|)|[a-zA-Z0-9]+)/';
+
     /**
      * {@inheritdoc}
      */
     public function downloadViaLink($locator, $key = 'Download', $headers = '')
     {
-        $url = $this->assertVisibleLink($locator)->getAttribute('href');
-        $baseUrlRegExp = '/^(http(s|):[\/]{2}|)(www\.|)[a-zA-Z0-9]+\.[a-zA-Z]+(\:[\d]+|)/';
+        $link = $this->assertVisibleLink($locator)->getAttribute('href');
 
-        if (!preg_match($baseUrlRegExp, $url)) {
+        if (!preg_match(self::$baseUrlRegExp, $link)) {
             $currentUrl = $this->getSession()->getCurrentUrl();
 
-            $baseFilter = preg_match(
-                $baseUrlRegExp,
-                $currentUrl,
-                $baseCurrentUrl
-            );
-
-            if ($baseFilter === 0) {
+            if (!preg_match(self::$baseUrlRegExp, $currentUrl, $linkParts)) {
                 throw new ExpectationException('Could not generate base url from "' . $currentUrl . '"');
             }
 
-            $url = $baseCurrentUrl[0] . $url;
+            // Checks if URL is relative.
+            if (strpos($link, '/' === 1)) {
+                // Append to base URL
+                $link = $linkParts[0] . $link;
+            } else {
+                // Resolve the relative URL to a fully qualified URL
+                $link = substr($currentUrl, 0, strpos($currentUrl, '/')) . $link;
+            }
         }
 
-        $this->download($url, $key, $headers);
+        $this->download($link, $key, $headers);
     }
 
     /**
--- a/src/Behat/FlexibleMink/Context/WebDownloadContext.php
+++ b/src/Behat/FlexibleMink/Context/WebDownloadContext.php
@@ -24,10 +24,22 @@ trait WebDownloadContext
     public function downloadViaLink($locator, $key = 'Download', $headers = '')
     {
         $url = $this->assertVisibleLink($locator)->getAttribute('href');
+        $baseUrlRegExp = '/^(http(s|):[\/]{2}|)(www\.|)[a-zA-Z0-9]+\.[a-zA-Z]+(\:[\d]+|)/';
 
-        // If the path does not contain a base_url, prepend the one specified in the configuration file.
-        if (!preg_match('/^(http(s|):[\/]{2}|)(www\.|)[a-zA-Z0-9]+\.[a-zA-Z]{3}/', $url)) {
-            $url = $this->getMinkParameter('base_url') . $url;
+        if (!preg_match($baseUrlRegExp, $url)) {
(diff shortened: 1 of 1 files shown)


src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -23,7 +23,14 @@
      */
     public function downloadViaLink($locator, $key = 'Download', $headers = '')
     {
-        $this->download($this->assertVisibleLink($locator)->getAttribute('href'), $key, $headers);
+        $url = $this->assertVisibleLink($locator)->getAttribute('href');
+
+        // If the path does not contain a base_url, prepend the one specified in the configuration file.
+        if (!preg_match('/^(http(s|):[\/]{2}|)(www\.|)[a-zA-Z0-9]+\.[a-zA-Z]{3}/', $url)) {
+            $url = $this->getMinkParameter('base_url') . $url;
+        }
```
Chekote: This won't work for a couple of reasons:

1. The base URL is not guaranteed to be only a host name. It can also include part of the path. E.g. if you're base URL was `http://www.stdcheck.com/manage` and you used this method to download a link that was `/manage/requisitions/1.pdf`, this method would try to download `http://www.stdcheck.com/manage/manage/requisitions/1.pdf`.

2. It is possible to navigate to a page on a different host when writing tests. As such, it's not safe to assume that the download link is for a file on the server referenced by the base_url.

To resolve both of these problems, you should use `$this->getSession()->getCurrentUrl()` and extract the protocol, hostname and port (if present) from that.

- [ ] Done
- [ ] Approved

Chekote:
Code review for 8b6cab5 is complete.

awesomeousPrime:
@Chekote Code review change has been made.

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -23,7 +23,26 @@
      */
     public function downloadViaLink($locator, $key = 'Download', $headers = '')
     {
-        $this->download($this->assertVisibleLink($locator)->getAttribute('href'), $key, $headers);
+        $url = $this->assertVisibleLink($locator)->getAttribute('href');
+        $baseUrlRegExp = '/^(http(s|):[\/]{2}|)(www\.|)[a-zA-Z0-9]+\.[a-zA-Z]+(\:[\d]+|)/';
```
Chekote: This should be a constant, as there's no need to create the string and assign it to a variable each time this function is invoked.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -23,7 +23,26 @@
      */
     public function downloadViaLink($locator, $key = 'Download', $headers = '')
     {
-        $this->download($this->assertVisibleLink($locator)->getAttribute('href'), $key, $headers);
+        $url = $this->assertVisibleLink($locator)->getAttribute('href');
```
Chekote: It's confusing having so many variables in a single function have variations on the name url. I suggest changing this to `$link`.

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -23,7 +23,26 @@
...
+        $baseUrlRegExp = '/^(http(s|):[\/]{2}|)(www\.|)[a-zA-Z0-9]+\.[a-zA-Z]+(\:[\d]+|)/';
+
+        if (!preg_match($baseUrlRegExp, $url)) {
+            $currentUrl = $this->getSession()->getCurrentUrl();
+
+            $baseFilter = preg_match(
+                $baseUrlRegExp,
+                $currentUrl,
+                $baseCurrentUrl
```
Chekote: I suggest changing this to `$linkParts`, for the same reason mentioned [here](https://github.com/Medology/FlexibleMink/pull/35/files#r66816334), and that this is the typical naming convention for the 3rd parameter to a preg_match.

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -23,7 +23,26 @@
...
+            );
+
+            if ($baseFilter === 0) {
+                throw new ExpectationException('Could not generate base url from "' . $currentUrl . '"');
+            }
```
Chekote: I don't think it's necessary to store the result of the preg_match in a variable, and using a var name like "filter" seems a little odd. I'd suggest putting the preg_match in the if statement directly:

```php
if (!preg_match($baseUrlRegExp, $currentUrl, $baseCurrentUrl)) {
  throw new ExpectationException('Could not generate base url from "' . $currentUrl . '"');
}
```

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -23,7 +23,26 @@
...
+
+            $url = $baseCurrentUrl[0] . $url;
```
Chekote: One other scenario where this will not currently work is if the url is relative.

For example, say that you're in `http://www.mysite.com/folder` and you click a link to download `files/one.pdf` (no leading slash). This *should* resolve to `http://www.mysite.com/folder/files/one.pdf`, but since we're not checking for a leading slash in the download url, this will currently resolve to `http://www.mysite.comfiles/one.pdf`.

Not only is the path missing the `folder` component, but the top level domain has changed to `comfiles` because of the missing slash.

To fix this, you'll need to check if the `url` has a leading slash and treat it differently if so:

```php
if (strpos($url, '/') === 1) {
  $url = $baseCurrentUrl[0] . $url;
} else {
  $url = substr($currentUrl, 0, strrpos($currentUrl, '/')) . $url;
}
```

- [ ] Done
- [ ] Approved

Chekote:
Testing code review for 8e9eefb is complete.

awesomeousPrime:
@Chekote code review changes have been made.  Also, change to regular expression to account for subdomains other than `www` and items without dots like localhost for example.

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -18,12 +18,35 @@
...
+            // Checks if URL is relative.
+            if (strpos($link, '/' === 1)) {
+                // Append to base URL
+                $link = $linkParts[0] . $link;
+            } else {
+                // Resolve the relative URL to a fully qualified URL
+                $link = substr($currentUrl, 0, strpos($currentUrl, '/')) . $link;
+            }
+
+            $link = $linkParts[0] . $link;
```
Chekote: I don't believe this should still be here. It's overriding what's happening with the if statement above.

Chekote:
Testing code review for d018f7f is complete.

awesomeousPrime:
@Chekote Change has been made.

Chekote:
14f8fab 👍

PR #40, opened 2016-06-21
adding assertion for javascript single value

awesomeousPrime:
Adding an assertion to check for a single javascript value.

### DIFF
--- a/src/Behat/FlexibleMink/Context/JavaScriptContext.php
+++ b/src/Behat/FlexibleMink/Context/JavaScriptContext.php
@@ -103,4 +103,23 @@ trait JavaScriptContext
 
         return $value;
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @Then the javascript variable :variableName should have the value of :expectedValue
+     */
+    public function assertJavascriptVariable($variableName, $expectedValue)
+    {
+        $returnedValue = $this->getSession()->evaluateScript(
+            'return ' . $variableName . ';'
+        );
+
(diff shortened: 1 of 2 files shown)


src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -103,4 +103,24 @@ protected function getRawOrJson($value)
...
+
+    /**
+     * {@inheritdoc}
+     *
+     * @Then the javascript variable :variableName should have the value of :expectedValue
+     */
+    public function assertJavascriptVariable($variableName, $expectedValue)
+    {
+        $returnedJsonData = $this->getSession()->evaluateScript(
+            'return JSON.stringify(' . $variableName . ');'
```
RedEnchilada: Is there a reason to JSON-encode the variable and then decode it after returning? You should be able to just return the variable as-is, and the driver should handle returning the proper variable type.

RedEnchilada:
@awesomeousPrime Code review complete. If I'm wrong about the type handling and it does need to be handled that way, then this is good to merge as-is.

src/Behat/FlexibleMink/Context/JavaScriptContext.php
awesomeousPrime: Yes this is not needed.  I've gone ahead and changed this.

awesomeousPrime:
@RedEnchilada code review change has been made.

RedEnchilada:
👍

PR #42, opened 2016-06-30
Store converts spaces to snake case when injecting stored values.

Chekote:
Requirement:
To be able to write feature files where the injected values are referred to in plain english, instead of
their snake case. e.g.

Before:
`Then I should see (the requisition_number of the order)`

After:
`Then I should see (the requisition number of the order)` or `Then I should see (the Requisition Number of the Order)`


RedEnchilada:
👌

PR #43, opened 2016-07-05
Store can be initialized via onStoreInitialized()

Chekote:
The store resets the registry at the beginning of each scenario, but
I have a requirement where my Scenario's need access to something in
the store at all times.

To facilitate this, I added a check for an onStoreInitialized method
after the registry is cleared, and if it exists, the StoreContext will
call it to populate the newly cleared registry.


awesomeousPrime:
@Chekote 👍

PR #44, opened 2016-07-07
Fix float casting only capturing whole number.

Chekote:
The `@Transform` for floats regex had two capturing groups, which caused only the first to be passed to the transformer. This meant that only the digits to the left of the decimal place were used when converting the string to a float. e.g. 1.99 became 1.

I removed the capturing groups from the regex as they were not needed. Without them, the entire string is passed to the transformer.


johnnie502:
👍

PR #51, opened 2016-08-02
Image Load Checks

TaysirTayyab:
# Overview
To ensure proper front-end coverage, developers should be able to test if the correct image loaded or not. This PR adds a check to ensure that an image was loaded correctly (to test for broken images).

# Changes
## New WebDownloadContext
- New method `checkImageLoaded` which uses the DOM properties of `naturalHeight` and `naturalWidth` to see if the loaded image is actually there
- New Meth ([I'm not in danger, I *am* the danger!](https://www.youtube.com/watch?v=9JS-iK0lD0s))

## New Tests
- New test to check if FlexibleMink functions can determine if an image has or has not loaded correctly
- New web files to support testing
- New assets directory for images in web


medology-bugman:

web/image-load-test.html
```diff
@@ -0,0 +1,22 @@
+<html>
+
+<head>
+  <style>
+    img {
+      width: 427px;
+      height: 120px;
```
atroppe: There's no need to specify both width and height, and generally we wouldn't since it could cause warping of the image.  I realize this won't affect the test since it's only checking to make sure neither are "0".  It's just more clear what's happening if you remove one of them.

features/WebDownloadContext.feature
```diff
@@ -0,0 +1,13 @@
...
+  I should have assertions to check if assets downloaded
+
+  Background:
+    Given I am on "/image-load-test.html"
+
+  Scenario: Developer Can Test if an Image Link is Valid (Loads)
+    Then I should see "img/medology.png" image in "valid-image"
+
+  Scenario: Developer Can Test if an Image Link is Invalid (Broken)
+    Then I should not see an image in "invalid-image"
```
atroppe: So you're checking to make sure the image is broken?  Is this step just to confirm that the test actually works if the image is broken?

web/image-load-test.html
```diff
@@ -0,0 +1,22 @@
...
+    }
+  </style>
+  <title>Table Test for Flexible Mink</title>
+</head>
+
+<body>
+  <h1>Valid Image Link</h1>
+  <img src="img/medology.png" id="valid-image" />
+  <h1>Invalid (Broken) Image</h1>
+  <img src="img/doesnotexist.png" id="invalid-image" />
```
atroppe: Since the test is meant to be used with the data-qa-id attribute, how were you able to test this without adding it?

src/Behat/FlexibleMink/PseudoInterface/WebDownloadContextInterface.php
```diff
@@ -24,4 +26,13 @@
      * @param string $headersString Headers to pass with the curl request.
      */
     abstract public function download($file, $key = 'Download', $headersString = '');
+
+    /**
+     * This method checks if the image for an <img> tag actually loaded.
+     *
+     * @param  string                   $xpath The xpath of the <img> tag to check
+     * @throws ElementNotFoundException If an <img> tag was not found at {@paramref $xpath}
```
atroppe: this is a nice addition

features/bootstrap/FeatureContext.php
```diff
@@ -16,4 +21,60 @@ public function putStoreStep($key, TableNode $attributes)
...
+
+        if (!$image) {
+            throw new ExpectationException("Expected an img tag with id '$locator'. Found none!", $session);
+        }
+
+        if ($image->getAttribute('src') != $imgSrc) {
+            throw new ExpectationException("Expected src '$imgSrc'. Instead got '" . $image->getAttribute('src') . "'.", $session);
+        }
+
+        if (!$this->checkImageLoaded($image->getXpath())) {
```
awesomeousPrime: This should probably use `waitFor`

features/bootstrap/FeatureContext.php
```diff
@@ -16,4 +21,60 @@ public function putStoreStep($key, TableNode $attributes)
...
+     */
+    public function assertImageNotLoaded($locator)
+    {
+        $image = $this->getSession()->getPage()->find('css', "img#$locator");
+
+        if (!$image) {
+            throw new ExpectationException("Expected an img tag with id '$locator'. Found none!", $this->getSession());
+        }
+
+        if ($this->checkImageLoaded($image->getXpath())) {
```
awesomeousPrime: Same thing mentioned [here](https://github.com/Medology/FlexibleMink/pull/51#discussion_r73387061)

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -74,4 +76,34 @@ public function download($file, $key = 'Download', $headersString = '')
...
+  complete: document.evaluate("{$xpath}", document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.complete,
+  height: document.evaluate("{$xpath}", document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.naturalHeight,
+  width: document.evaluate("{$xpath}", document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.naturalWidth
+}
+JS;
+
+        $imgProperties = $driver->evaluateScript($script);
+
+        $isLoaded = true;
+        if (!$imgProperties['complete'] || $imgProperties['width'] === 0 || $imgProperties['height'] === 0) {
```
jose-ochoa: The if statement could be reduced into line 102 using ternary operator

- [ ] Done
- [ ] Approved

web/image-load-test.html
```diff
@@ -0,0 +1,22 @@
+<html>
```
jose-ochoa: Please add doctype

- [ ] Done
- [ ] Approved

features/bootstrap/FeatureContext.php
awesomeousPrime: It seems this should use `waitFor` incase the image takes a while to load

- [ ] Done
- [ ] Approved

web/image-load-test.html
```diff
@@ -0,0 +1,22 @@
...
+      padding: 5px;
+      border: 2px solid #ccc;
+    }
+  </style>
+  <title>Table Test for Flexible Mink</title>
+</head>
+
+<body>
+  <h1>Valid Image Link</h1>
+  <img src="img/medology.png" id="valid-image" />
```
jose-ochoa: Please add alt tags to have valid element also on line 19

- [ ] Done
- [ ] Approved

features/bootstrap/FeatureContext.php
awesomeousPrime: Same thing mentioned [here](https://github.com/Medology/FlexibleMink/pull/51#discussion_r73388327)

jose-ochoa:
Back end code review complete. 1d7abe0

awesomeousPrime:
@TaysirTayyab testing review complete 1d7abe0

features/WebDownloadContext.feature
TaysirTayyab: Yeah this is just a negative check to make sure that if the image is broken, the test properly detects that it is.

web/image-load-test.html
TaysirTayyab: The method added to Flexible Mink is `checkImageLoaded` which takes an xpath to the image. The FeatureContext here only looks at the id, but when we add this to stdcheck, then I will use the xpath from the appropriate QA field (which will use the `data-qa-id`).

features/bootstrap/FeatureContext.php
TaysirTayyab: I added the `waitFor` around the `WebDownloadContext\checkImageLoaded` instead.

features/bootstrap/FeatureContext.php
TaysirTayyab: See https://github.com/Medology/FlexibleMink/pull/51#discussion_r73424912

web/image-load-test.html
jose-ochoa: XHTML why??

jose-ochoa:
Back end code review complete d671f6a
👍

web/image-load-test.html
TaysirTayyab: Why not?

awesomeousPrime:
@TaysirTayyab testing looks good to me

!reassign testing @Chekote

medology-bugman:

src/Behat/FlexibleMink/Context/WebDownloadContext.php
```diff
@@ -80,4 +82,33 @@ public function download($file, $key = 'Download', $headersString = '')
...
+    width: document.evaluate("{$xpath}", document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.naturalWidth
+}
+JS;
+
+            $imgProperties = $driver->evaluateScript($script);
+
+            return $imgProperties['complete'] ? $imgProperties : false;
+        });
+
+        return $result['width'] !== 0 && $result['height'] !== 0;
```
awesomeousPrime: I took another look at this and it seems that the waitFor would only reattempt when there is no image in the DOM but not when the image is not fully loaded so this may not work. 

This method is intended to return false whenever the image has not been fully loaded but waitFor relies on Exceptions.  

It might be a better place to place the waitFor within the assert functions.

Let me know what your thoughts are

- [ ] Done
- [ ] Approved

TaysirTayyab: The FeatureContext is not a part of flexible mink, but a context for testing flexible mink, so it definitely needs to go here.

The complete property on the img reports if the image is finished loading. If complete is true, then the image is loaded (or failed to load with timeout, 404 or whatever). Thus the waitFor looks at complete, and if the image is finished loading, it sends back the properties, otherwise it keeps sending false forcing the waitFor to keep running.

TaysirTayyab:
!reassign testing

medology-bugman:

RedEnchilada:
👍

PR #52, opened 2016-08-03
Method Scoping Change in TableContext

TaysirTayyab:
# Problem

Sometimes, as a developer, I need to write custom methods to check rows/cells in a table. As such, I need access to the parsed table from the FlexibleMink/TableContext.

# Fix
The scope of the following two methods is changed to protected to allow them to be accessed by any contexts which inherit Flexible Mink.


TaysirTayyab:
!force code review

medology-bugman:

goodevilgenius:
back-end 👍

PR #53, opened 2016-08-03
Extracting part that generates full URL into it's own method so that …

awesomeousPrime:
Extracting part of method which generates full URL of an item. 

For example: 

```html
<a href="/testing.php">Testing</a>
<!-- Not sure if the site below is real this is just for demonstration purposes...Not responsible for any material encountered -->
<a href="https://www.flexiblemink.com/testing>Testing</a>
```

With this, both of these href attributes would generate a full URL in the case of the first example, the current url would be used so if a person is on `https://www.stdcheck.com/adslkjfsl`

the url would be https://www.stdcheck.com/testing.php

The second example should just generate what is passed in as is since it's a full path already.

### DIFF
--- a/src/Behat/FlexibleMink/Context/WebDownloadContext.php
+++ b/src/Behat/FlexibleMink/Context/WebDownloadContext.php
@@ -25,8 +25,14 @@ trait WebDownloadContext
      */
     public function downloadViaLink($locator, $key = 'Download', $headers = '')
     {
-        $link = $this->assertVisibleLink($locator)->getAttribute('href');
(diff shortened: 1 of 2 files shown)


awesomeousPrime:
!force code review

medology-bugman:

RedEnchilada:
Your example is not syntactically valid because you're missing a quotation mark.

RedEnchilada:
👍  6577f08

awesomeousPrime:
submitting separate pull request for the changes mentioned in CR :P

PR #58, opened 2016-08-16
Test coverage

RedEnchilada:
Add tests for a lot of the functionality currently exposed by FlexibleMink. Not everything is covered yet, but a lot of commonly-used functionality now has tests ensuring that `Then`s both pass and fail (in all cases) properly, or that `When`s produce the expected action. (This means the Kanban story isn't quite complete yet, but I figured it'd be better to get early feedback in this process instead of bunching it all up into one monstrous PR after doing everything.)

A few fixes were also included to contexts where undesired behavior was discovered while writing tests.


medology-bugman:

behat.yml
bin/php

TaysirTayyab:
This is absolutely fantastic @RedEnchilada! Front-end review passed a888724 👍

features/StoreContext.feature
```diff
@@ -0,0 +1,47 @@
...
+     Then the assertion should pass
+
+    Examples:
+      | Value |
+      | 2     |
+      | 0     |
+      | 4.7   |
+      | true  |
+      | false |
+      | "car" |
```
johnnie502: We may need to assert if class instance could be stored and accessed

johnnie502:
Testing review complete, this is almost a comprehensive test set for flexible-mink, good job!

bin/php
```diff
@@ -5,4 +5,4 @@ for i in "$@"; do
   ARGS="$ARGS ${i}"
 done
 
-docker run -it -v $(pwd):/app -w /app php:cli php $ARGS
+docker run -it -v $(pwd):/app --net flexiblemink_default --link chrome --rm -w /app php:cli php $ARGS
```
Chekote: I believe you can access all containers in the network without specifying the link argument.

features/AlertContext.feature
```diff
@@ -0,0 +1,43 @@
...
+      | confirm |
+      | cancel  |
+
+  Scenario: Alert assertion fails properly when no alert exists
+    Given I am on "index.html"
+     When I assert that I should see an alert containing "alert text"
+     Then the assertion should throw an ExpectationException
+      And the assertion should fail with the message "No alert is open"
+
+  Scenario: Alert assertion fails properly when no alert exists
```
Chekote: This Scenario name doesn't seem correct. The same file is being opened on line 7 of this Feature and the assertion is passing. So an alert does exist. It just does not match the specified text.

- [ ] Done
- [ ] Approved

features/AlertContext.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Alert context
+  In order to ensure that JavaScript alerts work as expected
+  As a developer
+  I need to be able to assert various states of alerts
+
+  Scenario Outline: Developer can test when alerts appear
+    Given I am on "alert.html"
```
Chekote: If we encapsulate this in a step that describes the intended behavior, then this Scenario would be easier to read and not quite so "coincidental":

```cucumber
Scenario Outline: Developer can test when alerts appear
  Given there is an alert containing "alert text"
   When I assert that I should see an alert containing "alert text"
  ...
```

- [ ] Done
- [ ] Approved

features/AlertContext.feature
```diff
@@ -0,0 +1,43 @@
...
+     When I <Action> the alert
+     Then I should see "<Action> selected"
+
+    Examples:
+      | Action  |
+      | confirm |
+      | cancel  |
+
+  Scenario: Alert assertion fails properly when no alert exists
+    Given I am on "index.html"
```
Chekote: Same thing here with describing the intended state:

```cucumber
Scenario Outline: Alert assertion fails properly when no alert exists
  Given there is not an alert containing "alert text"
   When I assert that I should see an alert containing "alert text"
  ...
```

- [ ] Done
- [ ] Approved

features/AlertContext.feature
```diff
@@ -0,0 +1,43 @@
...
+     When I assert that I should see an alert containing "alert text"
+     Then the assertion should throw an ExpectationException
+      And the assertion should fail with the message "No alert is open"
+
+  Scenario: Alert assertion fails properly when no alert exists
+    Given I am on "alert.html"
```
Chekote: Same thing here with describing the intended state:

```cucumber
Scenario Outline: Alert assertion fails properly when no alert exists
  Given there is an alert containing "alert text"
   When I assert that I should see an alert containing "alert butt"
  ...
```

- [ ] Done
- [ ] Approved

features/AlertContext.feature
```diff
@@ -0,0 +1,43 @@
...
+     When I assert that I should see an alert containing "alert butt"
+     Then the assertion should throw an ExpectationException
+      And the assertion should fail with the message "Text 'alert butt' not found in alert"
+          # Selenium bombs out if we leave an alert up at the end of a test
+     When I confirm the alert
```
Chekote: You should be able to remove this and instead add the following to the context:

```php
    /**
     * Restarts the browser session.
     *
     * @BeforeScenario
     */
    public function restartSession()
    {
        $this->getSession()->restart();
    }
```

- [ ] Done
- [ ] Approved

features/AlertContext.feature
```diff
@@ -0,0 +1,43 @@
...
+
+  Scenario Outline: Developer can test the results of filling in prompts
+    Given I am on "prompt.html"
```
Chekote: Same thing here with describing the intended state:

```cucumber
Scenario Outline: Developer can test the results of filling in prompts
  Given there is a prompt containing "prompt?"
   When I assert that I should see a prompt containing "prompt?"
  ...
```

- [ ] Done
- [ ] Approved

Chekote:
Back end review for a888724 is complete.

Great work on filling out the tests for this project!

features/AlertContext.feature
RedEnchilada: I added an `@AfterScenario` hook to `AlertContext` that closes any open alerts instead. It has an additional benefit of preventing the whole testing suite from bombing out if another project using FlexibleMink runs into an unexpected alert.

features/StoreContext.feature
RedEnchilada: There's no code in `StoreContext` to handle classes in particular, so what would be tested?

johnnie502: All right, seems it doesn't matter then.

features/AlertContext.feature
Chekote: Does the `$this->getSession()->restart();` not work if there are alerts present?

Chekote:
Assuming that it is not possible to restart a session if there are alerts (see [this comment](https://github.com/medology/flexiblemink/pull/58#discussion_r75300210)), then this is good to merge.

3dfc7cb 👍

johnnie502:
👍  Testing

features/AlertContext.feature
RedEnchilada: I went with closing the alert for performance reasons. Using `$this->getSession()->restart();` takes the tests from 0m49.95s to 1m34.12s (almost a 1s increase per scenario), which seems unreasonable for tests that gain nothing else out of the session reboot. Using`$this->getSession()->reset();` instead doesn't solve the issue of alerts staying open, so it still bombs out.

PR #60, opened 2016-08-17
Fix parallel runner not distributing properly after scenario outlines

RedEnchilada:
**Method:**
- Run a set of tests with a scenario outline on more nodes than the outline has examples

**Expected Behavior:**
- All of the tests should run, distributed properly among the nodes
- Line numbers for examples should be accurate in case of test failure

**Actual Result:**
- Tests after the outline aren't all run, and some are run on multiple nodes
- The line numbers for examples were incorrect

**Cause:**
- The scenario count was not incremented if no examples were being run for an outline on a particular node, because an exception was thrown before the increment call
- The original array used the line numbers as the key for each example, and the filter was using array_values which discarded them

**Solution:**
Rewrote `filterExampleNode` to preserve array keys (to fix the line issue) and step through each example while continually incrementing the counter (which fixes the distribution issue)

### DIFF
--- a/src/Behat/ParallelWorker/Filter/ParallelWorkerFilter.php
+++ b/src/Behat/ParallelWorker/Filter/ParallelWorkerFilter.php
@@ -32,28 +32,22 @@ class ParallelWorkerFilter extends SimpleFilter
      */
     private function filterExampleNode(ExampleTableNode $examples)
     {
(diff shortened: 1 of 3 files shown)


medology-bugman:

awesomeousPrime:
@RedEnchilada back-end review complete 👍

PR #62, opened 2016-08-23
Assert Field Contains with Waitfor

joshmedeski:
Adds a wait for to the `assertFieldsContains` function for scenarios if Javascript changes the value of an input.

For https://github.com/Medology/healthlabs.com/pull/154

### DIFF
--- a/features/assertFieldContains.feature
+++ b/features/assertFieldContains.feature
@@ -1,0 +1,9 @@
+Feature: Assert Field Contains
+  In order to test content inside a field
+  As a developer
+  I should have a waitfor on the assertion
+
+ Scenario: Developer can test for asserting a field that is updated with a delay
+    When I am on "/assert-field-contains.html"
+    When I press "Fill in delayed field with a delay"
+    Then the "Delayed field" field should contain "delayed value"
--- a/features/assertFieldContains.feature
+++ b/features/assertFieldContains.feature
@@ -5,5 +5,5 @@ Feature: Assert Field Contains
 
  Scenario: Developer can test for asserting a field that is updated with a delay
     When I am on "/assert-field-contains.html"
-    When I press "Fill in delayed field with a delay"
+     And I press "Fill in delayed field with a delay"
     Then the "Delayed field" field should contain "delayed value"
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -41,6 +41,19 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      */
+    public function assertFieldContains($field, $value)
+    {
+        $field = $this->fixStepArgument($field);
+        $value = $this->fixStepArgument($value);
+
+        $this->waitFor(function () use ($field, $value) {
+            parent::assertFieldContains($field, $value);
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function assertPageContainsText($text)
     {
         $text = $this->injectStoredValues($text);
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -43,9 +43,6 @@ class FlexibleContext extends MinkContext
      */
     public function assertFieldContains($field, $value)
     {
-        $field = $this->fixStepArgument($field);
-        $value = $this->fixStepArgument($value);
-
         $this->waitFor(function () use ($field, $value) {
             parent::assertFieldContains($field, $value);
         });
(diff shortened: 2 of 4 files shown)


joshmedeski:
!force code review

medology-bugman:

features/assertFieldContains.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Assert Field Contains
+  In order to test content inside a field
+  As a developer
+  I should have a waitfor on the assertion
+
+ Scenario: Developer can test for asserting a field that is updated with a delay
+    When I am on "/assert-field-contains.html"
+    When I press "Fill in delayed field with a delay"
```
johnnie502: Please change `When` to `And`

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -41,6 +41,19 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      */
+    public function assertFieldContains($field, $value)
+    {
+        $field = $this->fixStepArgument($field);
+        $value = $this->fixStepArgument($value);
```
johnnie502: Please also add `$value = $this->injectStoredValues($value);` so we can assert the stored object value. like: `Then the "Delayed field" field should contain "(the name of the User)"`

johnnie502:
Also please add assertion to assert the false scenario can be assert fail like https://github.com/Medology/FlexibleMink/blob/master/features/ContainerContext.feature

johnnie502:
Testing review complete https://github.com/Medology/FlexibleMink/pull/62/commits/8c77bec4de14dd7622041ce5cf9046135e464475 Flexible mink is getting better everyday

src/Behat/FlexibleMink/Context/FlexibleContext.php
joshmedeski: I believe this is unnecessary extra work because this is just extending the default Behat behavior. Asserting stored object is dealt by the original function.

johnnie502: I see the problem here, you are right, we can probably add that later when we actually need it.

johnnie502:
Testing 👍  !reassign testing

medology-bugman:

web/assert-field-contains.html
```diff
@@ -0,0 +1,22 @@
...
+    <script>
+        function fillDelayedFieldWithDelay() {
+            setTimeout(function() {
+                document.getElementById("delayed-field").value = "delayed value"
+            }, 2000);
+        }
+    </script>
+</form>
+</body>
+</html>
```
awesomeousPrime: Please add a blank line at the end of the file.

- [ ] Done
- [ ] Approved

awesomeousPrime:
@joshmedeski testing review complete 6444182

awesomeousPrime:
@joshmedeski testing review complete 👍  3715e3f

web/assert-field-contains.html
```diff
@@ -0,0 +1,23 @@
...
+<head>
+    <meta charset="UTF-8">
+    <title>Assert Field Contains Delayed Value</title>
+</head>
+<body>
+<form>
+    <h1>Assert Field Contains Delayed Value</h1>
+    <label for="delayed-field">Delayed field</label>
+    <input type="text" id="delayed-field">
+    <button type="button" onclick="fillDelayedFieldWithDelay()">Fill in delayed field with a delay</button>
```
atroppe: fill delayed field with delay... lol. clever way of testing this.

src/Behat/FlexibleMink/Context/FlexibleContext.php
TaysirTayyab: Also FlexibleMink doesn't have the `Entity` context.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -41,6 +41,19 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      */
+    public function assertFieldContains($field, $value)
+    {
+        $field = $this->fixStepArgument($field);
```
TaysirTayyab: We don't need to fix `$fields` and `$values` here since the base method `parent::assertFieldContains` handles this already.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
```diff
@@ -26,6 +26,14 @@
     abstract public function assertPageContainsText($text);
 
     /**
+     * This method overrides the MinkContext::assertPageContainsText() default behavior for assertFieldContains to
+     * ensure that it waits for the text to be available with a max time limit.
+     *
+     * @see MinkContext::assertFieldContains
+     */
+    abstract public function assertFieldContains($field, $value);
```
TaysirTayyab: Since nothing is catching the exception that bubbles up from the assert down the call stack, please add `@throws ExpectationException` here.

- [ ] Done
- [ ] Approved

TaysirTayyab:
back-end review complete 8c77bec

TaysirTayyab:
back-end review passed fbf5b7f 👍

PR #68, opened 2016-09-10
Gherkin Regex Syntax Error and CircleCI Being a Pain

TaysirTayyab:
### Problem
`Statement Parsing`
The Gherkin statement for checking if a particular cell in a table was not parsing correctly by the IDE. Instead of showing a valid statement as being valid, the IDE instead showed it as being invalid.

screen shot 2016-09-10 at 12 24 48 pm

### Docker Dependency Failure
Also `docker-compose` randomly started failing on CircleCI. So fixing that too. So I can run tests again.

screen shot 2016-09-10 at 4 13 23 pm

# Cause
### Statement Parsing
In `TableContext::assertCellValue`, the regex for the Gherkin statement was malformed in referencing the variables match groups were to be assigned to.

As a result, the statement would *parse* correctly during runtime, but the IDE failed to recognize the statement as valid.

### Docker Dependency Failure
I honestly have no clue why this shit randomly blew up.

# Fix
### Statement Parsing
The regex statements in `TableContext::assertCellValue` were fixed and the IDE now properly recognizes the statements as valid.

### Docker Dependency Failure
Updated `circle.yml` to install the right version of the python six dependancy.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,6 +11,7 @@ machine:
 dependencies:
   override:
     - sudo pip install docker-compose
+    - sudo pip install 'six==1.10.0'
     - docker login -e ${DOCKER_EMAIL} -u ${DOCKER_USERNAME} -p ${DOCKER_PASSWORD}
     - docker-compose -p flexiblemink -f docker/docker-compose.yml up -d
     - sleep 5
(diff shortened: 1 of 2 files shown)


medology-bugman:

circle.yml

johnnie502:
👍  Back-end

PR #69, opened 2016-09-15
fix the anchor tag without href is not detected

johnnie502:
This fix the problem that Flexible mink can't detect anchor tag `<a>` when there is no `href` in the attribute.

_It is currently blocking the stdcheck my account redesign story_

PR #71, opened 2016-09-19
Fix assertVisibleLink not Checking Title/Id/Rel/Alt

TaysirTayyab:
# Overview

As a part of #69, @johnnie502 updated FlexibleMink to also detect anchor tags which don't have the `href` attribute (as this is quite common in Angular). Unfortunately the implementation only allowed matching on text content of the anchor tag. The default behavior of Mink allowed matching on the `id`, `rel`, `title`, `text()`, and `alt` (for images inside `a`) properties.

# Changes

To allow a clean port of this method into stdcheck, I'm updating the `assertVisibleLink` to use the named `link` selector, but just without the `href`.

It's hacky, since it modifies the default link selector's xpath via regex, but this is because the `replacements` and `selectors` properties used in the `NameSelector` are private and cannot be accessed. Maybe there's a way around that, but I didn't see it.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -171,28 +171,25 @@ class FlexibleContext extends MinkContext
     {
         $locator = $this->fixStepArgument($locator);
 
-        $links = $this->getSession()->getPage()->findAll(
-            'named',
-            ['content', $this->getSession()->getSelectorsHandler()->xpathLiteral($locator)]
-        );
+        // the link selector in Behat/Min/src/Selector/NamedSelector requires anchor tags have href
+        // we don't want that, because some don't, so rip out that section. Ideally we would load our own
+        // selector with registerNamedXpath, but I want to re-use the link named selector so we're doing it
+        // this way
+        $xpath = $this->getSession()->getSelectorsHandler()->selectorToXpath('named', ['link', $locator]);
+        $xpath = preg_replace('/\[\.\/@href\]/', '', $xpath);
 
-        /** @var NodeElement $link */
-        foreach ($links as $link) {
-            if ($link->getTagName() != 'a') {
-                continue;
-            }
-            try {
-                $visible = $link->isVisible();
-            } catch (UnsupportedDriverActionException $e) {
-                return $link;
-            }
+        /** @var NodeElement[] $links */
+        $links = array_filter($this->getSession()->getPage()->findAll('xpath', $xpath), function ($link) {
+            /* @var NodeElement $link */
+            return $link->isVisible();
+        });
 
-            if ($visible) {
-                return $link;
-            }
+        if (empty($links)) {
+            throw new ExpectationException("No visible link found for '$locator'", $this->getSession());
         }
 
-        throw new ExpectationException("No visible link found for '$locator'", $this->getSession());
+        // $links is NOT numerically indexed, so just grab the first element and send it back
+        return array_shift($links);
     }
 
     /**
--- a/features/FlexibleContext/AssertLinkClickable.feature
+++ b/features/FlexibleContext/AssertLinkClickable.feature
@@ -6,13 +6,39 @@ Feature: Assert Link is clickable
   Background:
     Given I am on "/link.html"
 
-  Scenario: Step passes if link is clickable
-    Then I follow "I am a link"
+  Scenario: When Multiple Matches are Found, First Visible Match is Used
+    When I follow "Link"
 
-  Scenario: Step passes if link without href
-    Then I follow "an anchor tag without href"
+  Scenario Outline: Visible Links are Clickable
+    Then I follow "<locator>"
 
-  Scenario: Step fails if link is invisible
-    When I assert that I follow "an invisible anchor tag"
+    Examples:
+      | locator                   |
+      | Visible Text Link w/ Href |
+      | Visible Text Link no Href |
+      | id-visible-href           |
+      | id-visible-nohref         |
+      | title-visible-href        |
+      | title-visible-nohref      |
+      | rel-visible-href          |
+      | rel-visible-nohref        |
+      | alt-visible-href          |
+      | alt-visible-nohref        |
+
+  Scenario Outline: Trying to Click Invisible Link Throws Exception
+    When I assert that I follow "<locator>"
     Then the assertion should throw an ExpectationException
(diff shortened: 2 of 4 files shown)


TaysirTayyab:
!force review

TaysirTayyab:
!force code review

medology-bugman:

awesomeousPrime:
@TaysirTayyab testing review 👍 

!reassign testing

medology-bugman:

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -171,28 +171,23 @@ public function assertVisibleLink($locator)
...
+        // we don't want that, because some don't, so rip out that section. Ideally we would load our own
+        // selector with registerNamedXpath, but I want to re-use the link named selector so we're doing it
+        // this way
+        $xpath = $this->getSession()->getSelectorsHandler()->selectorToXpath('named', ['link', $locator]);
+        $xpath = preg_replace('/\[\.\/@href\]/', '', $xpath);
+
+        /** @var NodeElement[] $links */
+        $links = $this->getSession()->getPage()->findAll('xpath', $xpath);
+
+        if (count($links) > 1) {
```
johnnie502: @TaysirTayyab Thank you so much to make the fix for this issue! I must overlooked that...
The intention of this assertion is to `Finds the first matching visible link on the page.`, so I suggest not to throw the exception if links are more than 1, but throw it when **visible links are more than 1**.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -171,28 +171,23 @@ public function assertVisibleLink($locator)
...
+            throw new ExpectationException("Multiple matches found for '$locator'", $this->getSession());
```
goodevilgenius: I'm not sure I see a reason to throw an error if there are more than one.

It says, `@Given the :locator link is visible`, not `@Given one :locator link is visible`.

All previous versions of this method assert that there was at least one link that matched, and returned that one, so this change would not be backwards compatible.

src/Behat/FlexibleMink/Context/FlexibleContext.php
RedEnchilada: Actually, why was anything past the `findAll` changed to begin with? I get that ideally you'd only be naming unique links in tests, but this limitation makes it extremely difficult to test legacy code.

atroppe:
Front end seems pretty thorough! 👍

src/Behat/FlexibleMink/Context/FlexibleContext.php
TaysirTayyab: 👍

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -171,28 +171,21 @@ public function assertVisibleLink($locator)
...
+        // selector with registerNamedXpath, but I want to re-use the link named selector so we're doing it
+        // this way
+        $xpath = $this->getSession()->getSelectorsHandler()->selectorToXpath('named', ['link', $locator]);
+        $xpath = preg_replace('/\[\.\/@href\]/', '', $xpath);
+
+        /** @var NodeElement[] $links */
+        $links = $this->getSession()->getPage()->findAll('xpath', $xpath);
+
+        if (!$links || !$links[0]->isVisible()) {
+            throw new ExpectationException("No visible link found for '$locator'", $this->getSession());
```
goodevilgenius: This only asserts that the first found link is visible.

What you really need is to check that at least one link is visible, and return that one.

In other words, if the first link is invisible, but the second (or nth where `n>1`) one is visible, the test should still pass.

So ...:

```php
foreach($links as $link) {
  if ($link->isVisible()) {
    return $link;
  }
}
throw new ExpectationException("No visible link found for '$locator'", $this->getSession());
```

TaysirTayyab: Removed `findAll` and now we're just looking for the first one, which is the same behavior as `MinkContext::clickLink`.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -171,28 +171,21 @@ public function assertVisibleLink($locator)
...
+        // we don't want that, because some don't, so rip out that section. Ideally we would load our own
+        // selector with registerNamedXpath, but I want to re-use the link named selector so we're doing it
+        // this way
+        $xpath = $this->getSession()->getSelectorsHandler()->selectorToXpath('named', ['link', $locator]);
+        $xpath = preg_replace('/\[\.\/@href\]/', '', $xpath);
+
+        /** @var NodeElement[] $links */
+        $link = $this->getSession()->getPage()->find('xpath', $xpath);
+
+        if (!$link || !$link->isVisible()) {
```
RedEnchilada: The problem with this is it only works if the first link it finds is visible. If an invisible link precedes a visible link in the page, this will throw an exception instead of going to the first visible link to click.

I presume you're trying to optimize the `foreach` loop out of this function, but until Mink itself has a way to filter for visible elements only when finding elements, there's really no way to get around having that in place.

TaysirTayyab: I'm trying to replicate the behavior of `MinkContext::clickLink`. What that does, is it uses `findAll` on the selector, then takes the first element of the result and clicks it.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -171,28 +171,21 @@ public function assertVisibleLink($locator)
...
+            throw new ExpectationException("No visible link found for '$locator'", $this->getSession());
```
goodevilgenius: This still doesn't quite work. `$this->getSession()->getPage()->find` should get the first element, visible or not. You're still only testing that the first matching node is visible, not that at least one is visible.

TaysirTayyab: See my previous response to @RedEnchilada.

TaysirTayyab:
Also the method here is misnamed a bit. @johnnie502's original PR (#69) was to make links without href's clickable. So the visible thing is just a bonus, the point of the method was to make sure that href's are clickable.

RedEnchilada:
pls apply the StyleCI patch

TaysirTayyab:
@goodevilgenius @RedEnchilada @awesomeousPrime @atroppe Thanks for the quick turn-around guys!

PR #72, opened 2016-09-20
Assert page contains texts

goodevilgenius:
As part of the better lab redesign, I added a method to `WebContext` that @johnnie502 [suggested was more appropriate in FlexbileMink](https://github.com/Medology/stdcheck.com/pull/1581#pullrequestreview-763451).

This new method can be used to assert that multiple strings occur in a single page, regardless of the order, in a single step, rather than specifying thus:

```
Then I should see "string1"
 And I should see "string2"
 And I should see "string3"
```

With this new method, it can be specified this way:

```
Then I should see the following:
  | string1 |
  | string2 |
  | string3 |
```

This is a little easier to read, and useful if you have already specified a list of data earlier in a scenario.

### DIFF
--- a/features/FlexibleContext/assertTexts.feature
+++ b/features/FlexibleContext/assertTexts.feature
@@ -12,19 +12,19 @@ Feature: Assert Page Contains Texts
       | Line two   |
       | Line three |
 
-    Scenario: Step injects values properly
-      Given the following is stored as "list":
-          | first_entry  | Line one   |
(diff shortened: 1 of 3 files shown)


medology-bugman:

Chekote:
Great quality-of-life improvement! 👍

jose-ochoa:
Back-end code review complete. e96518b 👍

PR #70, opened 2016-09-16
Add Hook To Modify Entity in StoreContext::injectStoredValue

TaysirTayyab:
# Overview
Due to some issues with Laravel and Model relationships, we need a way to reload a model that's stored in the key store. This PR creates the infrastructure to allow us to hook into the `injectStoredValue` method to manipulate the entity before accessing the property of it.

# Changes
```
bin/phpunit
bin/composer
```
- these are now version locked to a particular image of phpunit and composer

```
composer.json
composer.lock
```
- adding phpunit and all it's dependencies to the dev environment

```
StoreContext.php
StoreContextInterface.php
```
- adding a `callable` to the `injectStoredValues` method to serve as "hook" for the entity

```
phpunit.xml
StoreContextTest.php
```
- new unit tests for `StoreContext::injectStoredValues`

PR #73, opened 2016-09-20
Fixing InjectStoredValues to Handle Magic Properties

TaysirTayyab:
# Overview

With the new hook for `StoreContext::injectStoredValues` (#70), the method now throws exceptions when attempting to access magic properties (ie. properties set via `Class::__get`). This fix updates the method to handle these magic properties.

# Cause
PHP's `property_exists` only applies to explicit properties, thus will not detect magic properties.

# Changes
```
StoreContext.php
StoreContextTest.php
```
- `injectStoredValues` now updated to handle magic properties using `isset` instead of `property_exists`. This requires that `Class::__isset` is defined, which is standard for dynamic properties.

```
bin/phpunit
```
- Since `phpunit` is now installed by composer, its not a good idea to run it in a separate container with separate source. The script now uses the installed phpunit in the php container

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -123,8 +123,8 @@ trait StoreContext
             }
 
             if (
-                (is_object($thing) && !property_exists($thing, $thingProperty)) ||
-                (is_array($thing) && !isset($thing, $thingProperty))
+                (is_object($thing) && !isset($thing->$thingProperty)) ||
+                (is_array($thing) && !isset($thing[$thingProperty]))
             ) {
                 throw new Exception("$thingName does not have a $thingProperty property");
             }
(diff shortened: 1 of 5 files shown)


medology-bugman:

bin/phpunit
phpunit.xml

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -122,10 +122,7 @@ protected function injectStoredValues($string, callable $onGetFn = null)
                 throw new Exception('The $onGetFn method must return an object or an array!');
             }
 
-            if (
-                (is_object($thing) && !property_exists($thing, $thingProperty)) ||
-                (is_array($thing) && !isset($thing, $thingProperty))
-            ) {
+            if ((is_object($thing) || is_array($thing)) && !isset($thing->$thingProperty)) {
```
awesomeousPrime: This seems to allow an array with the if `is_array()`. With that in mind, wouldn't any time an array is used, `isset($thing->$thingProperty)` return false?

Is that intended?

TaysirTayyab: Ahh shit yes good catch.

awesomeousPrime:
@TaysirTayyab back-end 👍 52cc60a

TaysirTayyab:
!reassign backend @awesomeousPrime

PR #76, opened 2016-09-20
Add nth Key Parsing to isStored Method

TaysirTayyab:
# Overview

The `StoreContext::get` method parses the provided `$key` to separate out a potential "n" value and key name. Essentially, you can provide keys in the form of `nth key` and have the method parse the `n` and the `key` out of it.

This behavior is unfortunately not used in the `isStored` method. So if you want to check that a given key is stored before you retrieve it, you have to the key yourself.

# Changes
This PR creates a method for key parsing. Any (local) member needing a key parsed can call the method and have the parse done. This is implemented in `get` and `isStored` so that you can now check the same key that you plan on getting without having to parse stuff yourself.

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -66,13 +66,30 @@ trait StoreContext
     }
 
     /**
+     * Converts a key of the form "nth thing" into "n" and "thing".
+     *
+     * @param  string $key The key to parse
+     * @return array  For a key "nth thing", returns [thing, n], else [thing, null]
+     */
+    private function parseKey($key)
+    {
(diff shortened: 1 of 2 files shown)


src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -66,13 +66,30 @@ protected function assertIsStored($key, $nth = null)
...
+        return [$key, $nth];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function get($key, $nth = null)
+    {
+        if (!$nth) {
+            list($key, $nth) = $this->parseKey($key);
```
Chekote: Duples! nice 👍

medology-bugman:

PR #77, opened 2016-09-21
Moved phpunit to dev dependencies

Chekote:
__Problem__
PHPUnit was declared as a regular dependency, which meant that any
client consuming this composer package would end up getting phpunit
installed. More importantly, the version requirements for phpunit
imposed by this package would have to be adhered to by clients that
consumed this package.

This causes particular issues with projects that require stable
dependencies, because the phpunit version required by this package
is not stable.

__Fix__
I moved the phpunit dependency to the dev dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,11 +7,11 @@
     "php": ">=5.4.41",
     "behat/behat": "^3.0",
     "symfony/config": "~2.8",
-    "behat/mink-extension": "^2.0",
-    "phpunit/phpunit": "5.5.*"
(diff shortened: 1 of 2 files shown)


medology-bugman:

composer.json
composer.lock

Chekote:
!reassign back-end @goodevilgenius

medology-bugman:

goodevilgenius:
That did it. Now `composer update medology/flexible-mink` properly updates to the most recent version.

PR #81, opened 2016-10-14
Adding Several Prototypes to Mink Interface

TaysirTayyab:
The following prototypes are added to the Mink psuedo-interface.

- `fillFields()`
- `selectOption()`

These are to resolve Scrutinizer issues showing up in projects which use Flex Mink.

### DIFF
--- a/src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
+++ b/src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
@@ -2,11 +2,12 @@
 
 use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Session;
+use Behat\MinkExtension\Context\MinkContext;
 
 /**
  * Pseudo interface for tracking the methods of the MinkContext.
  *
- * @see \Behat\MinkExtension\Context\MinkContext.
+ * @see MinkContext.
  */
 trait MinkContextInterface
 {
--- a/src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
+++ b/src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
@@ -1,11 +1,12 @@
 <?php namespace Behat\FlexibleMink\PseudoInterface;
 
+use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Session;
 
 /**
  * Pseudo interface for tracking the methods of the MinkContext.
  *
- * @see Behat\MinkExtension\Context\MinkContext.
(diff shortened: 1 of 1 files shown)


medology-bugman:

src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
```diff
@@ -1,11 +1,12 @@
 <?php namespace Behat\FlexibleMink\PseudoInterface;
 
+use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Session;
 
 /**
  * Pseudo interface for tracking the methods of the MinkContext.
  *
- * @see Behat\MinkExtension\Context\MinkContext.
+ * @see \Behat\MinkExtension\Context\MinkContext.
```
Chekote: Please add this to the use section and use the classes short name here.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
```diff
@@ -32,6 +33,25 @@
     abstract public function fillField($field, $value);
 
     /**
+     * Fills in form fields with provided table.
+     *
+     * @param TableNode $fields Pairs of fields and values to fill. Example:
+     *                          [
+     *                          [ username, bruceWayne]
+     *                          [ password, iLoveBats123]
+     *                          ]
```
Chekote: Since this is a TableNode, this example should be documented using the cucumber table node syntax, as opposed to the current array syntax:

```cucumber
| username | bruceWayne   |
| password | iLoveBats123 |
```

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/PseudoInterface/MinkContextInterface.php
TaysirTayyab: I had it this way because PHPStorm still complains even when it's added via use.

screen shot 2016-10-17 at 10 46 17 am

Chekote: That's because you're still not using the classes short name. PHPStorm is complaining because both your `use` and your `see` are using the fully qualified names, and therefore the `use` is unnecessary. If you use the classes short name, PHPStorm will not complain:

screenshot 2016-10-18 08 52 56

TaysirTayyab: Ahhh it wooooorks!

PR #82, opened 2016-10-18
Allow custom logic for property exists.

Chekote:
Problem:
I need to use injectStoredValues to inject the deleted_at from an Eloquent model, whatever that value might be. However, deleted_at will be set to null if the model has not been deleted. This is a perfectly valid value, but because injectStoredValues uses isset to determine if the item from the store has a property, it throws an Exception claiming that the model does not have a deleted_at property, when it clearly does (it's just null).

Fix:
After trying a myriad of different options, the only solution that was close to satisfactory was to add the option to pass a lambda into the injectStoredValues function that allows you to change the behavior of how a thing it determined to have a value or not. This allows an Eloquent user to pass in a lambda that supports Eloquent, while avoiding tightly coupling the FlexibleMink framework to Eloquent itself.

For example, I could pass in the following:

```php
   function eloquentHasProperty($thing, $property) {
        if ($thing instanceof Model) {
            return array_key_exists($property, $thing->getAttributes()) || $thing->hasGetMutator($property);
        } else {
            return
                (is_object($thing) && !isset($thing->$property)) ||
                (is_array($thing) && !isset($thing[$property]));
        }
    }
```

Which would allow injectStoredValues to properly detect properties on an Eloquent model, even if they were null, or generated via a mutator.

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -116,10 +116,22 @@ trait StoreContext
     /**
      * {@inheritdoc}
      */
-    protected function injectStoredValues($string, callable $onGetFn = null)
+    protected function injectStoredValues($string, callable $onGetFn = null, callable $hasValue = null)
     {
         if ($onGetFn && (new ReflectionFunction($onGetFn))->getNumberOfParameters() != 1) {
             throw new Exception('Method $onGetFn must take one argument!');
+        }
+
+        if ($hasValue) {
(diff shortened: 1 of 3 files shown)


medology-bugman:

Chekote:
!reassign testing @TaysirTayyab

medology-bugman:
Paging @TaysirTayyab to review the testing work in this pull request!

billmedology:
Amazing solution @Chekote! 👍

awesomeousPrime:
@Chekote back-end review 👍 a0c5fd3

PR #83, opened 2016-10-21
Change waitFor to Sleep for 1/10 Second

TaysirTayyab:
### Overview
Simple PR, changes the waitFor function to sleep for a 1/10th of a second instead of a full second (since that's actually quite long).

We're doing this to catch any kind of transitions which happen in a window of less than 1 second (such as hide/show status messages).

### Miscellaneous
I'm also updating the `composer.json` with meta information about the project.

### DIFF
--- a/src/Behat/FlexibleMink/Context/SpinnerContext.php
+++ b/src/Behat/FlexibleMink/Context/SpinnerContext.php
@@ -26,8 +26,8 @@ trait SpinnerContext
                 $lastException = $e;
             }
 
-            # sleep for 10^8 nanoseconds (0.1 second)
-            time_nanosleep(0, pow(10,8));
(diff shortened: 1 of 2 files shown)


TaysirTayyab:
!force code review

medology-bugman:

composer.json

src/Behat/FlexibleMink/Context/SpinnerContext.php
```diff
@@ -26,7 +26,8 @@ public function waitFor(callable $lambda, $timeout = 30)
                 $lastException = $e;
             }
 
-            sleep(1);
+            // sleep for 10^8 nanoseconds (0.1 second)
+            time_nanosleep(0, pow(10, 8));
```
johnnie502: This is a good change, the devil is in the details!

PR #84, opened 2016-10-21
adding waitForPageNotContainsText method

awesomeousPrime:
There is currently already a method which can already test for this. The only difference is the exception message thrown.  Since the step definition is a when, this exception message seems more clear.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -100,15 +100,21 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      *
-     * @When I wait until I do not see :text
+     * @When I wait until I see and then do not see :text
      */
     public function waitForPageNotContainsText($text)
     {
+        $this->waitFor(function () use ($text) {
+            parent::assertPageContainsText($text);
+        }, 15);
+
         try {
-            $this->assertPageNotContainsText($text);
+            $this->waitFor(function () use ($text) {
+                parent::assertPageContainsText($text);
+            }, 15);
         } catch (ExpectationException $e) {
             throw new ResponseTextException(
-                "Wait For timed out waiting for '$text' to no longer appear.", $this->getSession()
+                "Timed out waiting for '$text' to no longer appear.", $this->getSession()
             );
         }
     }
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -6,6 +6,7 @@ use Behat\FlexibleMink\PseudoInterface\FlexibleContextInterface;
 use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Element\NodeElement;
 use Behat\Mink\Exception\ExpectationException;
+use Behat\Mink\Exception\ResponseTextException;
 use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Behat\MinkExtension\Context\MinkContext;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


medology-bugman:

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -98,6 +99,22 @@ public function assertPageNotContainsText($text)
 
     /**
      * {@inheritdoc}
+     *
+     * @When I wait until I do not see :text
```
TaysirTayyab: Instead of creating a new function for, this, why not just add the Gherkin to `assertPageNotContainsText`?

Also to be honest, I don't see how this is cleared. Could you provide a use case where this statement would be cleared than "Then I should not see text"?

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -98,6 +99,22 @@ public function assertPageNotContainsText($text)
...
+     */
+    public function waitForPageNotContainsText($text)
```
TaysirTayyab: In stdcheck `WebContext`, the `assertPageNotContainsText` is wrapped in a wait for. I would go ahead and overload the Mink `assertPageNotContainsText` method here with the `waitFor`  and then we can remove it from `WebContext`.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/FlexibleContext.php
awesomeousPrime: There was an issue in stdcheck manage where I had to check for the text "Loading" to appear on the page and then wait until it no longer appeared so the steps looked like this: 

```cucumber
Then I should see "Loading"
 And I should not see "Loading"
```

In [code review](https://github.com/Medology/stdcheck.com/pull/1919#discussion_r84513761) it was mentioned that this would be more readable to have it as 
`When I wait until I do not see "Loading"`

The reason for the new method was just to have a different error message than the one that `assertPageNotContainsText` throws. Since it's a `@When` step it seemed that the default message of `The text "shouldNotAppear" appears in the text of this page, but it should not` was kind of odd.  Functionally it's the same thing though and it's possible to just add the pattern here to `assertPageNotContains` Text but it also seemed that assert should have `@Then` steps though o.o

I may just be over thinking it and can go ahead and add the pattern into `assertPageNotContainsText` let me know what your thoughts are.

src/Behat/FlexibleMink/Context/FlexibleContext.php
awesomeousPrime: It seems that `FlexibleContext` already has an `assertPageNotContainsText` method that has the wait for so we should already be able to remove it from `WebContext` on stdcheck.  I'll go ahead and create another PR for that.

awesomeousPrime:
@TaysirTayyab comment has been responded to.

src/Behat/FlexibleMink/Context/FlexibleContext.php
TaysirTayyab: Okay I understand the why, so now my concern is the how. This doesn't actually check to see the text was there to begin with. For a `When` check (something like `When I no longer see "Loading"`) the assertion should be that we see it followed by an assertion that we no longer see it.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/FlexibleContext.php
awesomeousPrime: This is removed in [PR](https://github.com/Medology/stdcheck.com/pull/1974)

awesomeousPrime:
@TaysirTayyab code review changes have been made

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -98,6 +99,28 @@ public function assertPageNotContainsText($text)
...
+            parent::assertPageContainsText($text);
+        }, 15);
+
+        try {
+            $this->waitFor(function () use ($text) {
+                parent::assertPageContainsText($text);
+            }, 15);
+        } catch (ExpectationException $e) {
+            throw new ResponseTextException(
+                "Wait For timed out waiting for '$text' to no longer appear.", $this->getSession()
```
TaysirTayyab: Can probably just remote the `Wait For` part here and say `Timed out waiting ...`.

PR #85, opened 2016-10-27
Allow colspan attribute

jose-ochoa:
While the table context doesn't support the `colspan` attribute on `td` or `th` it should allow it if they are set to `colspan="1"` and throw an exception if the number of columns they span is greater than 1.

This was done to support 3rd party JavaScript libraries that generate the table DOM setting the `colspan=1` attribute.

### DIFF
--- a/src/Behat/FlexibleMink/Context/TableContext.php
+++ b/src/Behat/FlexibleMink/Context/TableContext.php
@@ -104,7 +104,7 @@ trait TableContext
             // finds all td|th elements that have a colspan tag. We do this because we don't want any HEAD rows
             // when some of the cells span multiple columns
             /** @var NodeElement[] $splitCell */
-            $splitCell = $row->findAll('xpath', '/*[@colspan and (self::td or self::th)]');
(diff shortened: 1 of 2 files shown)


jose-ochoa:
!force code review @johnnie502

medology-bugman:

src/Behat/FlexibleMink/Context/TableContext.php
```diff
@@ -104,7 +104,7 @@ private function findHeadColumns($table)
             // finds all td|th elements that have a colspan tag. We do this because we don't want any HEAD rows
             // when some of the cells span multiple columns
             /** @var NodeElement[] $splitCell */
-            $splitCell = $row->findAll('xpath', '/*[@colspan and (self::td or self::th)]');
+            $splitCell = $row->findAll('xpath', '/*[@colspan>\'1\' and (self::td or self::th)]');
```
Chekote: Please update the `@throws ElementNotFoundException` in the docblock, as it still states that it is thrown if all head rows have a colspan attribute at all.

- [ ] Done
- [ ] Approved

PR #86, opened 2016-10-27
Configures Project for PHP 5.4

TaysirTayyab:
# Overview

Project has a requirement of at least php 5.4, but contains dependencies and assets that need or are a part of php 5.6.

# Changes
- Downgrade tests and files to use php 5.4
- update composer with 5.4 valid packages


bin/php
```diff
@@ -4,4 +4,4 @@ docker run --rm -it --env-file "$(pwd)"/docker/.env \
     -v "$(pwd)":/app -w /app \
 	--link web:dockermachine.local \
 	--net flexiblemink_default \
-	php:7-cli php $@
+	php:5.6-cli php $@
```
jose-ochoa: Why no 5.4?

- [ ] Done
- [ ] Approved

TaysirTayyab: 5.4 is no longer supported by PHP and is past it's lifetime. As such, php doesn't publish docker images for anything less that 5.6.

medology-bugman:

.gitignore
.idea/php.xml
bin/composer
bin/containers
bin/php
composer.json
composer.lock

PR #89, opened 2016-11-10
Context for assert field exists and patiences for button/options

jose-ochoa:
3 Changes:

- Added waitFor to press button function
- Added waitFor to check option function
- Added context to assert field exists, by default the entire page is the context if not defined.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -4,8 +4,8 @@ namespace Behat\FlexibleMink\Context;
 
 use Behat\FlexibleMink\PseudoInterface\FlexibleContextInterface;
 use Behat\Gherkin\Node\TableNode;
-use Behat\Mink\Element\DocumentElement;
 use Behat\Mink\Element\NodeElement;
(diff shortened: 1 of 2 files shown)


medology-bugman:

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -264,18 +269,22 @@ public function assertVisibleOption($locator)
     /**
      * {@inheritdoc}
      */
-    public function assertFieldExists($fieldName)
+    public function assertFieldExists($fieldName, DocumentElement $context = null)
```
Chekote: I think `DocumentElement` is going to be too specific. The `DocumentElement` is specifically the entire HTML document, which doesn't work for this use case because we want to specify a narrower scope. All other elements within the DOM would be `NodeElement`s.

We do not want to restrict this to a `NodeElement` either, as it would also be too specific, because `DocumentElement` is not a `NodeElement`.

The common ancestor of both of these classes is `TraversableElement`, which has all of the methods that we leverage in the `assertFieldExists` method. So I would recommend changing this parameter type to that.

- [ ] Done
- [ ] Approved

PR #91, opened 2016-11-15
Fix Element not having property since it was not found

jose-ochoa:
# Issue 

A fatal error was being produced when a `<label>` element was found with the `for` attribute set without a corresponding element with the matching `id`. 

# Solution
Check if a corresponding element is found before adding it to the `fields` array.

### DIFF
--- a/features/FlexibleContext/assertFieldExists.feature
+++ b/features/FlexibleContext/assertFieldExists.feature
@@ -37,8 +37,8 @@ Feature:  Assert Fields exists
        | Pizza     |
        | Hamburger |
        | Hot Dog   |
-    Then I check "Pizza"
-    Then I check "Hamburger"
-    Then I check "Hot Dog"
+    When I check "Pizza"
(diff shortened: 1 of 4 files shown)


medology-bugman:

features/FlexibleContext/assertFieldExists.feature
```diff
@@ -0,0 +1,44 @@
...
+       | Sushi |
+    Then the assertion should throw an ExpectationException
+     And the assertion should fail with the message "No input label 'Sushi' found"
+
+  Scenario: Developer Can Test for option with varying input/label setup
+    Then I should see the following fields:
+       | Pizza     |
+       | Hamburger |
+       | Hot Dog   |
+    Then I check "Pizza"
```
RedEnchilada: Should these be `When`s instead of `Then`s?

awesomeousPrime:
@jose-ochoa back-end 👍 96519cc

PR #98, opened 2017-01-10
adding waitFor for clickLink method

awesomeousPrime:
Adding a wait for on the clickLink method.


medology-bugman:

PR #99, opened 2017-01-12
Add waitFor to assertButtonDisabled

jose-ochoa:
In the instance where a button becomes enabled/disabled as part of DOM manipulation via JS, if the function to change disable state is slow due to performance the wait will prevent it from immediately falling.


medology-bugman:

PR #100, opened 2017-01-28
force rebase to origin on circleci

johnnie502:
This is for https://github.com/Medology/stdcheck.com/issues/2039

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,8 +10,9 @@ machine:
 
 dependencies:
   override:
+    - git config --global user.email "ci@medology.com" && git config --global user.name "Circle CI" && git rebase origin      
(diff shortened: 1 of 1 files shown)


medology-bugman:

circle.yml

circle.yml
```diff
@@ -10,8 +10,9 @@ machine:
 
 dependencies:
   override:
+    - git config --global user.email "ci@medology.com" && git config --global user.name "Circle CI" && git rebase origin
```
RedEnchilada: You might need to `git rebase origin master` to make sure it's rebasing against the proper branch. I think

johnnie502: The remote default branch is defaulted to master for rebased. But it is good to be explicit for those details, thanks for your suggestion.

PR #103, opened 2017-02-27
add radio button support and tests

johnnie502:
Add the Radio Button support, since there is no native support from mink project. (You can use the same one for dropdown menu though, which is not ideal)

### DIFF
--- a/features/FlexibleContext/assertRadioButton.feature
+++ b/features/FlexibleContext/assertRadioButton.feature
@@ -1,0 +1,58 @@
+Feature:  Assert Radio Button
+  In order to ensure that radio button elements behave as expected
+  As a developer
+  I should have radio button behavior assertions
+
+  Background:
+    Given I am on "/radio-buttons.html"
+
+  Scenario: Assert the radio button checked
+    Then the "Checked" radio button should be checked
+
+  Scenario: Assert the radio button not checked
+    Then the "Not Checked" radio button should not be checked
+
+  Scenario: Assertion fails reliably if checked radio button is expected as unchecked
+    When I assert that the "Checked" radio button should not be checked
+    Then the assertion should throw an ExpectationException
+     And the assertion should fail with the message 'Radio button "Checked" is checked, but it should not be.'
+
+  Scenario: Assertion fails reliably if unchecked checkbox is expected as checked
+    When I assert that the "Not Checked" radio button should be checked
+    Then the assertion should throw an ExpectationException
+     And the assertion should fail with the message 'Radio button "Not Checked" is not checked, but it should be.'
+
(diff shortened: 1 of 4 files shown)


medology-bugman:

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -592,4 +593,99 @@ public function assertCheckboxNotChecked($checkbox)
...
+
+            $radioButtons = array_filter($radioButtons, function (NodeElement $radio) {
+                return $radio->isVisible();
+            });
+
+            if (!$radioButtons) {
+                throw new ExpectationException('No Visible Radio Button was found on the page', $this->getSession());
+            }
+
+            usort($radioButtons, [$this, 'compareElementsByCoords']);
```
RedEnchilada: Since the sort callback used here fails if the driver doesn't support coordinate reads, I think you should do either of the following:

- Wrap this in a try/catch block that specifically catches the `UnsupportedDriverActionException` and ignores it
- Adjust the callback to silently handle the unsupported driver action

The former would probably be better practice.

- [ ] Done
- [ ] Approved

features/FlexibleContext/assertRadioButton.feature
```diff
@@ -0,0 +1,58 @@
...
+      And the "Radio Button 2" radio button should be checked
+      And the "Radio Button 3" radio button should not be checked
+     When I check radio button "Radio Button 3"
+     Then the "Radio Button 1" radio button should not be checked
+      And the "Radio Button 2" radio button should not be checked
+      And the "Radio Button 3" radio button should be checked
+     When I check radio button "Radio Button 1"
+     Then the "Radio Button 1" radio button should be checked
+      And the "Radio Button 2" radio button should not be checked
+      And the "Radio Button 3" radio button should not be checked
```
RedEnchilada: I don't think this scenario is necessary. The limitation of one radio box in a group being checked at a time is a browser feature and not something that FlexibleMink handles.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/FlexibleContext.php
RedEnchilada: The solution implemented does not solve the original problem, as the callback throws an exception before it reaches the point that was changed. You could replace throwing the exception with the return you added, although since `true` doesn't explicitly match the `int` return type you'd probably wanna return `-1` (to ensure the current order is kept, as sorting for a `0` return is undefined and thus not guaranteed to keep the same order).

johnnie502: I misunderstood it, I will make the change. Thanks!

PR #109, opened 2017-05-17
add assertion for options in select

johnnie502:
This will add assertion to assert that a form select(dropdown) control has the exact option items in the exact order listed in feature file.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -447,6 +447,58 @@ class FlexibleContext extends MinkContext
     }
 
     /**
+     * {@inheritdoc}
+     *
+     * @Then /^the "(?P<select>[^"]*)" select should only have the following option(?:|s):$/
+     */
+    public function assertSelectContainsExactOptions($select, TableNode $tableNode)
+    {
+        if (count($tableNode->getRow(0)) > 1) {
+            throw new InvalidArgumentException('Arguments must be a single-column list of items');
+        }
+
+        $expectedOptTexts = array_map([$this, 'injectStoredValues'], $tableNode->getColumn(0));
+
+        $select = $this->fixStepArgument($select);
+        $select = $this->injectStoredValues($select);
+        $selectField = $this->assertFieldExists($select);
+        $actualOpts = $selectField->findAll('xpath', '//option');
+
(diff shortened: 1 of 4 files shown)


medology-bugman:

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -447,6 +447,38 @@ public function assertSelectContainsOption($select, $existence, $option)
...
+            throw new ExpectationException('No option found in the select', $this->getSession());
+        }
+
+        $actualOptTexts = array_map(function ($actualOpt) {
+            /* @var NodeElement $actualOpt */
+            return $actualOpt->getText();
+        }, $actualOpts);
+
+        if ($actualOptTexts != $expectedOptTexts) {
+            throw new ExpectationException('Options in the select not match', $this->getSession());
```
RedEnchilada: For the sake of helping developers troubleshoot tests, I'd suggest adding more descriptive error messages to this assertion. `Select has more options than expected`, `Select has fewer options than expected`, and `Expected select option $option, but got $option` would probably be enough to provide useful errors.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -447,6 +447,38 @@ public function assertSelectContainsOption($select, $existence, $option)
     }
 
     /**
+     * {@inheritdoc}
+     *
+     * @Then /^the "(?P<select>[^"]*)" select should only have the following option(?:|s):$/
```
RedEnchilada: This can be written as `@Then the :select select should only have the following option(s):` and achieve the same flexibility, if you'd prefer that over the regex declaration.

johnnie502: Thanks for the suggestion, I would like to keep it as is since it is more clear when reference in feature file.

src/Behat/FlexibleMink/Context/FlexibleContext.php
johnnie502: I've update it but went another route since formatting could be tricky when printing the content of `$option`.

johnnie502:
!reassign front-end @joshmedeski

medology-bugman:

PR #110, opened 2017-05-30
Addition of overwrite to fill a field that is visible.

juanmrad:
This will make sure that the field that will be filled is a visible field by the user instead of selecting a fiend that might be hidden.


medology-bugman:

PR #111, opened 2017-06-14
Make the cancelAlerts() only executing when there's active session

johnnie502:
This is trying to improve the timeout problem that occurs frequently in other projects when tests in running in the selenium docker container.

### Cause
Becuase a `@AfterScenario` tag is added to the `cancelAlerts()` function, this function will be automatically executed after every scenario. Since there's no mechanism that behat can arrange the execution order of all `@AfterScneario` functions, when there is a function to clean the web driver session, and this function is executed before the `cancelAlerts()`, the `cancelAlerts()` will instantiate a new session and try to cancel the alerts. This newly created session will not be cleaned and it will be kept in WebDrive process across the test suite execution. Which slows down the chrome drive and causes the timeout.

### Implementation
* Disable the auto hook, or
* Clean the alert only if there's active session. https://github.com/Medology/FlexibleMink/pull/111#discussion_r121997591

### DIFF
--- a/src/Behat/FlexibleMink/Context/AlertContext.php
+++ b/src/Behat/FlexibleMink/Context/AlertContext.php
@@ -20,11 +20,15 @@ trait AlertContext
     /**
      * Clears out any alerts or prompts that may be open.
      *
-     * @AfterScenario @clearAlertsWhenFinished
+     * @AfterScenario
      * @Given there are no alerts on the page
      */
     public function clearAlerts()
     {
+        if (!$this->getSession()->getDriver()->isStarted()) {
+            return;
+        }
+
         try {
             $this->cancelAlert();
         } catch (NoAlertOpenError $e) {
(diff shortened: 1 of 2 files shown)


johnnie502:
!force code review

medology-bugman:

src/Behat/FlexibleMink/Context/AlertContext.php
```diff
@@ -20,7 +20,7 @@
     /**
      * Clears out any alerts or prompts that may be open.
      *
-     * @AfterScenario
+     * @AfterScenario @clearAlertsWhenFinished
```
Chekote: I like the original intent of this method in that it would automatically clear out alerts without the test developer having to explicitly call it.

You mentioned that the problem was additional sessions being created. Instead of adding an annotation to enable/disable this, we could modify the method to only attempt to clear the alerts if there is a WebDriver session currently active:

```php
public function clearAlerts()
{
  if (!$this->getSession()->getDriver()->isStarted()) {
    return;
  }

  ...
}
```

- [ ] Done
- [ ] Approved

PR #113, opened 2017-07-31
add spinner to the alert assertion

johnnie502:
Add a 15 seconds `waitFor` to the alert assertion

### DIFF
--- a/src/Behat/FlexibleMink/Context/AlertContext.php
+++ b/src/Behat/FlexibleMink/Context/AlertContext.php
@@ -63,15 +63,17 @@ trait AlertContext
      */
     public function assertAlertMessage($expected)
     {
-        try {
-            $actual = $this->getSession()->getDriver()->getWebDriverSession()->getAlert_text();
-        } catch (NoAlertOpenError $e) {
-            throw new ExpectationException('No alert is open', $this->getSession());
-        }
+        $this->waitFor(function () use ($expected) {
+            try {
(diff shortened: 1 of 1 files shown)


medology-bugman:

src/Behat/FlexibleMink/Context/AlertContext.php
```diff
@@ -63,15 +63,17 @@ public function cancelAlert()
...
+                throw new ExpectationException('No alert is open', $this->getSession());
+            }
 
-        if (strpos($actual, $expected) === false) {
-            throw new ExpectationException("Text '$expected' not found in alert", $this->getSession());
-        }
+            if (strpos($actual, $expected) === false) {
+                throw new ExpectationException("Text '$expected' not found in alert", $this->getSession());
+            }
+        }, 15);
```
Chekote: Why not use the default value for waitFors timeout?

johnnie502: 2 minutes seems too long. Or should we keep that as a standard?

Chekote: The long time is intended for highly fluctuating environments like CircleCI. The test will pass as soon as the alert shows up, so the long time is only going to take effect if the test is failing.

If 2mins is too long, then it would make more sense to change it across the board. I have concerns with the amount of times we are overriding this in tests, because it's increasing the number of places in the code where we need to worry about the timing issues. Having a single value (even a long one like 2 mins), gives us a single point where we can adjust the timing of these waitFors across the board.

PR #118, opened 2017-10-27
Update to new BDD standards

Chekote:
__Requirement__
We have defined new standards for our BDD tests that alleviate the need to have everything be a trait. FlexibleMink is the foundation for a lot of our tests, so for us to implement this standard on our projects that leverage it, we need to update it to these new standards.

__Implementation__
* The trait interfaces were removed, and their documentation moved to the implemeting context.
* The trait contexts were converted to classes, and stitched together where necessary using the more OOP gatherContexts approach
* The waitFor default timeout was changed from a default method arg to a static variable on the Spinner class that can be overridden
* The namespace structure was refactored to be beneath Medology. We don't own the top level Behat namespace, so we should not be creating classes there
* Given phrases that were being used on assert methods were removed (Given's are for ensure methods)
* The ContainerContext class was removed in favor of using the QualityAssurance approach that uses the data-qa-id. This is currently implemented in other projects, so will need to be extracted for sharing across projects.


medology-bugman:

.idea/php.xml
behat.yml
bin/php
composer.json
dictionary.dic
phpunit.xml

PR #123, opened 2017-11-13
Fix PHP image cannot go over Github API rate limit

Chekote:
Problem:
When hitting the Github API rate limit, the PHP image is asking for an auth token.

Cause:
The PHP image is not mounting the hosts ~/.composer directory. So even if the Docker host has an auth token, it is not used.

Fix:
I updated the PHP image to mount the ~/.composer directory. The new image also brings in some other quality of life features, such as the ability to ctrl-c out of the container, and run the script on both Linux and Mac.

PR #124, opened 2017-11-13
Update Composer bin to 1.5.2

Chekote:
Problem:
The composer script is actually an alias for a docker image that contains composer. Although this works, it is not necessary as we already have a PHP image in the project.

Solution:
In addition to updated the composer version, I modified the project so that the composer binary is not in version control. Instead it is installed via a command in init_project. The composer file that is installed is the normal phar, and is simply executed against the bin/php Docker image since the `#!/usr/bin/env php` at the beginning of the composer file will resolve to `./bin/php` if `./bin` is sufficiently high in the users PATH list (which it should be).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .idea/
-
+bin/composer
 /vendor/
 docker/.env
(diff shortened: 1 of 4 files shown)


medology-bugman:

.gitignore
bin/composer
bin/init/install_composer.sh
bin/init_project

Chekote:
This is dependent on https://github.com/Medology/FlexibleMink/pull/123. It will not pass until that is merged.

PR #125, opened 2017-11-16
Fixing issue with nested table check

awesomeousPrime:
When using the `TableContext::assertTableWithStructureExists()` method I noticed that I kept gettin an error whenever testing a table that had another table nested in it.

This was due to the method attempting to retrieve all rows and columns within a specified table instead of only checking for the most top level ones.


awesomeousPrime:
!force code review

medology-bugman:

PR #127, opened 2017-11-30
Remove retry from "Then" steps.

Chekote:
# Problem:
The waitFor functionality is implemented inconsistently. It should be applied uniformly to all "Then" steps.

Even if it is applied uniformly to all "Then" steps, it is still incumbent upon the developer to add the functionality to all new "Then" steps, and also incumbent upon the FlexibleMink consumer to add the functionality to their own "Then" steps. We would ideally like to remove the need for this repetitive plumbing.

# Implementation:
I wrote a Behat extension called [BehatRetryExtension](https://github.com/Chekote/BehatRetryExtension). This can be added to any project using Behat, and will automatically retry all "Then" steps that are attempted while the extension is active.

I initially added this extension to FlexibleMink, but after discussion with the team, we decided that the retry functionality should be left out of FlexibleMink, and the consumer of the FlexibleMink package can decide if they wish to bring in the BehatRetryExtension also.


medology-bugman:

composer.lock

Chekote:
!reassign front-end

medology-bugman:

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -36,36 +35,6 @@ class FlexibleContext extends MinkContext
...
-    public function assertPageAddress($page)
-    {
-        Spinner::waitFor(function () use ($page) {
-            parent::assertPageAddress($page);
-        });
-    }
-
     /**
      * This method overrides the MinkContext::assertPageContainsText() default behavior for assertPageContainsText to
      * ensure that it waits for the text to be available with a max time limit.
```
jose-ochoa: Should the description be updated?

- [ ] Addressed
- [ ] Approved

Chekote: Yes. Good catch. I've updated all docblocks that mention waiting.

PR #128, opened 2018-01-12
Add Check if Field is visible

juanmrad:
Addition of step definition to allow check if a field is visible or not when is it supposed to be visible or not.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -127,6 +127,26 @@ class FlexibleContext extends MinkContext
         }
     }
 
+
+    /**
+     * {@inheritdoc}
+     * @Then /^the field "(?P<field>[^"]+)" should(?P<not> not|) be visible$/
+     */
+    public function assertFieldIsVisible($field, $not){
+        $locator = $this->fixStepArgument($field);
+
+        $option = $this->getSession()->getPage()->find(
+          'named',
+          ['field', $this->getSession()->getSelectorsHandler()->xpathLiteral($locator)]
+        );
+
+        $shouldBeVisible = !$not;
+        if (($shouldBeVisible && !$option->isVisible()) || (!$shouldBeVisible && $option->isVisible())) {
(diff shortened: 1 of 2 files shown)


medology-bugman:

juanmrad:
!reassign testing

medology-bugman:

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
```diff
@@ -44,6 +44,13 @@
      */
     abstract public function assertPageAddress($page);
 
+    /**
+     * @param  string               $field The field to be checked
```
johnnie502: Please add a description for this function

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -127,6 +127,25 @@ public function assertPageContainsTextTemporarily($text)
         }
     }
 
+    /**
+     * {@inheritdoc}
+     * @Then /^the field "(?P<field>[^"]+)" should(?P<not> not|) be visible$/
+     */
+    public function assertFieldIsVisible($field, $not)
```
johnnie502: Since this step will assert the field is visible or not, `assertFieldVisibility` might be a better name.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -127,6 +127,25 @@ public function assertPageContainsTextTemporarily($text)
...
+    {
+        $locator = $this->fixStepArgument($field);
+
+        $option = $this->getSession()->getPage()->find(
```
johnnie502: Since this is going to locate a field, `$field` is a better name for this purpose.

johnnie502: Some of our pages (especially on manage) have fields that use the same name/label/locator, I would suggest `->findAll()` and throw exception if there's multiple fields being found by the same name. By only asserting the first matching fields, there's potential of creating bugs.

PR #130, opened 2018-01-24
Updated Noun-Store

Chekote:
Removed unused parseKey method and updated Noun Store to v2.

### DIFF
--- a/src/Medology/Behat/StoreContext.php
+++ b/src/Medology/Behat/StoreContext.php
@@ -3,8 +3,9 @@
 namespace Medology\Behat;
 
 use Behat\Behat\Context\Context;
+use Chekote\NounStore\Assert;
+use Chekote\NounStore\Store;
 use Exception;
-use nounstore\Store;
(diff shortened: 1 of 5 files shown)


medology-bugman:

composer.json
composer.lock

src/Medology/Behat/StoreContext.php
```diff
@@ -125,7 +118,7 @@ public function injectStoredValues($string, callable $onGetFn = null, callable $
             $thingName = $matches[2][$i];
             $thingProperty = str_replace(' ', '_', strtolower($matches[1][$i]));
 
-            if (!$this->has($thingName)) {
+            if (!$this->keyExists($thingName)) {
```
jose-ochoa: Shouldn't this be `$this->assert`?

- [ ] Addressed
- [ ] Approved

Chekote: Yes, that would be cleaner. Good catch.

PR #135, opened 2018-02-28
Fixing issue with asserting field exists

awesomeousPrime:
At the moment, it looks like `FlexibleContext::assertFieldExists()` attempts to find all fields that are named whichever field name is specified and returns the first visible one it finds when it exists.  If none is found through this way, it attempts to try to find fields by using label names as sort of a last resort.

When attempting to find items by the last resort method, `$context->find()` is used instead of `findAll` so it behaves differently in the last resort method compared to trying to find inputs by their name.

Because of this one can get into an issue where an item can have the same label name for multiple items, but only the first item found would be checked and returned. 

No issue exists for this it was just something that came up when working on another project.


medology-bugman:

PR #137, opened 2018-03-02
2.0

awesomeousPrime:
These are the same changes being made in the following PR to master for version 2.0 of this repo.

https://github.com/Medology/FlexibleMink/pull/135

### DIFF
--- a/web/assert-field-exists.html
+++ b/web/assert-field-exists.html
@@ -54,7 +54,6 @@
         }
 
         for (var i = 0; i < textElements.length; i++) {
-            console.log();
             text.push(textElements[i].getAttribute('name') + ': ' + textElements[i].value);
         }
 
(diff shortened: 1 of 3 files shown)


medology-bugman:

web/assert-field-exists.html
```diff
@@ -34,18 +34,32 @@ <h2>Favorite Foods</h2>
...
+
         var foods = [];
+        var text = [];
+
         for (var i = 0; i < checkedElements.length; i++) {
             foods.push(checkedElements[i].value);
         }
+
+        for (var i = 0; i < textElements.length; i++) {
+            console.log();
```
mehtaamish89: Do we need this `console.log()`? If yes, we need to add something for it to log?
- [ ] Added data to console.log.
- [ ] Removed the console.log.

awesomeousPrime: This is not needed I missed it.  Good catch

PR #143, opened 2018-05-08
Spin assertSelectContainsExactOptions()

Chekote:
Problem:
I need to be able to test for options in a select that are added via an asynchronous process. The method currently fails immediately if the options are not present.

Fix:
I modified the method to use a waitFor()


Chekote:
Wrong merge target.

PR #146, opened 2018-05-11
Addition of step and test for Partially match Tables.

juanmrad:
This will allow to find a table on the page without the need of `| * |` for empty values or values/columns that are not needed to be tested.

### DIFF
--- a/src/Behat/FlexibleMink/Context/TableContext.php
+++ b/src/Behat/FlexibleMink/Context/TableContext.php
@@ -483,4 +483,67 @@ trait TableContext
             );
         });
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @Then the table :name should contain the following values:
+     */
+    public function assertTableShouldContainTheFollowingValues($name, TableNode $tableNode)
+    {
+        $table = $this->getTableFromName($name);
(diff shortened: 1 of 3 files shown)


medology-bugman:

src/Behat/FlexibleMink/Context/TableContext.php
```diff
@@ -483,4 +483,67 @@ public function assertTableShouldHaveTheFollowingValues($name, TableNode $tableN
...
+        }
+    }
+
+    /**
+     * Checks if the expected row exists in the table provided and returns the row key where it was found.
+     *
+     * @param array $expectedRow The that is expected in the table.
+     * @param array $table       The table to find row
+     *
+     * @return int|bool False when the row was not found or the key where the row was found
```
goodevilgenius: Instead of returning `int|bool`, you could change it so that if the row isn't found, you return `-1`, so you only have to return `int`. This will be be beneficial if this library is upgraded to require PHP7, in which case you could add the return type.

If you do this, you probably want to change the method name to `getRowNumber` or something like that. And of course, change where it's called above to be `if (($key = $this->rowExistsInTable($row, $actual)) === -1)`

- [ ] Done
- [ ] Approved

PR #151, opened 2018-06-06
Change the regex for castStringToInt function

amberwjq:
It's for the behat test contains zipcode, so that we don't need to hack with single quote inside double quote  like "'07093'" anymore.


medology-bugman:

Chekote:
@amberwjq @joshmedeski @awesomeousPrime Please do not submit, approve or merge any changes for this project to the master branch. All PRs should be submitted to both the 1.0 and the 2.0 branches  (assuming they make sense for both). Also, 1.0 is deprecated, so all projects using it should plan to move to 2.0.

PR #152, opened 2018-06-12
Change regex for castStringToInit

amberwjq:
Same PR as https://github.com/Medology/FlexibleMink/pull/151
Instead of pushing to master branch, push it to 1.0 branch since healthlabs is using branch 1.0

**Problem I'm solving**
It's for the behat test contains zipcode, right now If I input "07093", the output will be "7093". And I have to hack with single quote inside double quote like "'07093'"

**How my change solves it.**

### DIFF
--- a/src/Behat/FlexibleMink/Context/TypeCaster.php
+++ b/src/Behat/FlexibleMink/Context/TypeCaster.php
@@ -10,7 +10,7 @@ trait TypeCaster
     /**
      * Casts a step argument from a string to an int.
      *
-     * @Transform /^((?!(0))[0-9]+)$/
+     * @Transform /^([1-9]\d*)$/
      * @param  string $string the string to cast.
      * @return int    The resulting int.
      */
--- a/src/Behat/FlexibleMink/Context/TypeCaster.php
+++ b/src/Behat/FlexibleMink/Context/TypeCaster.php
@@ -10,7 +10,7 @@ trait TypeCaster
     /**
      * Casts a step argument from a string to an int.
      *
-     * @Transform /^(\d+)$/
+     * @Transform /^((?!(0))[0-9]+)$/
      * @param  string $string the string to cast.
      * @return int    The resulting int.
      */


amberwjq:
!Force code-review

medology-bugman:

src/Behat/FlexibleMink/Context/TypeCaster.php
```diff
@@ -10,7 +10,7 @@
     /**
      * Casts a step argument from a string to an int.
      *
-     * @Transform /^(\d+)$/
+     * @Transform /^((?!(0))[0-9]+)$/
```
PrinceMegahit: `/^([1-9]\d*)$/` is also OK?

Chekote:
@amberwjq regarding your comment about master vs 1.0. This project should not have anything pushed to master. All changes should be submitted to both 1.0 and 2.0 (assuming they make sense for both). 1.0 is deprecated, so all projects using it should plan to move to 2.0.

src/Behat/FlexibleMink/Context/TypeCaster.php
Chekote: Why are we not allowing leading zeros? It is still an integer with a leading zero, and there are legitimate use cases where this is required.

Chekote:
Looks like my review message got cut off. Here's what it was supposed to say:

There is no explanation on this PR as to why this change was made. Please be sure that all of your PR's explain what problem you are solving, and how your change solves it.

src/Behat/FlexibleMink/Context/TypeCaster.php
amberwjq: @Chekote If allow leading zero, `intval($string)` function will cut off the leading zero. I've updated the PR explanation above to explain my problem.

Chekote: OK, that makes sense.

To answer your first question here: `/^([1-9]\d*)$/` would actually be better because it is explicitly saying that the first character has to be a 1-9 character, which is more (appropriately) restrictive than saying that it can be any character other than a zero. With the current pattern, "a1" would be converted to an int.

PR #153, opened 2018-06-14
Change regex for castStringToInit

amberwjq:
Same PR as https://github.com/Medology/FlexibleMink/pull/151
Instead of pushing to master branch, push it to both 1.0 and 2.0 branch  since all changes should be submitted to both 1.0 and 2.0 (assuming they make sense for both). 1.0 is deprecated, so all projects using it should plan to move to 2.0.

**Problem I'm solving**
It's for the behat test contains zipcode, right now If I input "07093", the output will be "7093". And I have to hack with single quote inside double quote like "'07093'"

**How my change solves it.**


medology-bugman:

PR #154, opened 2018-06-14
Fix Carbon instances are not formatted properly by injectStoredValues

Chekote:
# Problem:
Eloquent uses a static dateFormat property on the class which will cause the Carbon instances to be formatted when the model is converted to an array or JSON. If the Carbon instance is converted to a string via PHP, the dateFormat property is not going to be used. This can cause problems because comparing a string Carbon instance locally to one received from the server will result in different formatting.

# Fix:
I implemented a method that will determine if the property being read is a DateTime object, and if so, check for the existence of a dateFormat property on the host object. If one exists, the DateTime instance will be formatted as per the dateFormat property.

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -3,8 +3,11 @@
 namespace Behat\FlexibleMink\Context;
 
 use Behat\FlexibleMink\PseudoInterface\StoreContextInterface;
+use DateTime;
 use Exception;
+use ReflectionException;
 use ReflectionFunction;
+use ReflectionProperty;
 
 /**
  * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


medology-bugman:

Chekote:
!reassing testing @hopeseekr

Chekote:
!reassign back-end @hopeseekr

medology-bugman:
Paging @hopeseekr to review the back-end work in this pull request!

goodevilgenius:
!reassign testing @goodevilgenius

medology-bugman:
Paging @goodevilgenius to review the testing work in this pull request!

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -167,12 +170,46 @@ public function injectStoredValues($string, callable $onGetFn = null, callable $
...
+     *
+     * @param  mixed  $property the property to get from the object
+     * @param  object $thing    the object to get the value from
+     * @return mixed  the prepared value
+     */
+    protected function getValueForInjection($property, $thing)
+    {
+        $value = $thing->$property;
+
+        if ($value instanceof DateTime && is_object($thing)) {
```
goodevilgenius: To be completely safe, you might want to check if `$value instanceof DateTimeInterface` instead of `DateTime`, just on the random off chance that you get a `DateTimeImmutable` somewhere. (We had that issue in Raven once)

- [ ] Done
- [ ] Approved

Chekote: That's a good idea, and I would definitely like to implement it, but FlexibleMink supports PHP 5.4, which has the DateTime class, but not the DateTimeInterface. For me to implement this suggestion, I'd need to bump the requirement to PHP 5.5, which would be a major version change for FlexibleMink.

PR #155, opened 2018-06-15
Fix Carbon instances are not formatted properly by injectStoredValues

Chekote:
Implementation of #154 for 2.0 branch.

# Problem:
Eloquent uses a static dateFormat property on the class which will cause the Carbon instances to be formatted when the model is converted to an array or JSON. If the Carbon instance is converted to a string via PHP, the dateFormat property is not going to be used. This can cause problems because comparing a string Carbon instance locally to one received from the server will result in different formatting.

# Fix:
I implemented a method that will determine if the property being read is a DateTime object, and if so, check for the existence of a dateFormat property on the host object. If one exists, the DateTime instance will be formatted as per the dateFormat property.

### DIFF
--- a/src/Medology/Behat/StoreContext.php
+++ b/src/Medology/Behat/StoreContext.php
@@ -5,8 +5,11 @@ namespace Medology\Behat;
 use Behat\Behat\Context\Context;
 use Chekote\NounStore\Assert;
 use Chekote\NounStore\Store;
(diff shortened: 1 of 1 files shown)


medology-bugman:

Chekote:
!reassign testing

medology-bugman:

PR #156, opened 2018-06-18
Specify format when injecting values

Chekote:
# Problem:
The injectStoredValues method is simplistic in that it injects the stored value as is without any context. For the most part this is fine, but sometimes we have a model that has a DateTime which is being displayed by the presentation tier using a different format. For example, the created_at property of a model might include hours, mins and seconds, but the front-end may decide that they only wish to display year, month and day (probably in standard US format). If the stringified date from the store does not match how the front-end is presenting it, then it's not possible to say something like `I should see (the created at date of the order)`.

# Fix:
I implemented the ability to specify a format for the injected value. For example `I should see (the created date of the order as a US date)`. I added a few useful formats, and developers can easily add more as needed.

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -5,6 +5,7 @@ namespace Behat\FlexibleMink\Context;
 use Behat\FlexibleMink\PseudoInterface\StoreContextInterface;
 use DateTime;
 use Exception;
+use InvalidArgumentException;
 use ReflectionException;
 use ReflectionFunction;
 use ReflectionProperty;
@@ -19,6 +20,14 @@ trait StoreContext
 
     /** @var array */
     protected $registry;
+
+    protected static $FORMAT_MYSQL_DATE_AND_TIME = 'a MySQL date and time';
+    protected static $format_map = [
+        'a MySQL date'          => 'Y-m-d',
+        'a MySQL date and time' => 'Y-m-d H:i:s',
+        'a US date'             => 'm-d-Y',
+        'a US date and time'    => 'm-d-Y H:i:s',
+    ];
 
     /**
      * Clears the registry before each Scenario to free up memory and prevent access to stale data.
@@ -144,10 +153,13 @@ trait StoreContext
             };
         }
 
-        preg_match_all('/\(the ([^\)]+) of the ([^\)]+)\)/', $string, $matches);
+        preg_match_all('/\(the ([^\)]+) of the ([^\)]+?)( formatted as ([^\)]+))?\)/', $string, $matches);
         foreach ($matches[0] as $i => $match) {
             $thingName = $matches[2][$i];
             $thingProperty = str_replace(' ', '_', strtolower($matches[1][$i]));
+            $thingFormat = $matches[4][$i];
+
+            //dump($matches);
 
             if (!$this->isStored($thingName)) {
                 throw new Exception("Did not find $thingName in the store");
@@ -170,7 +182,11 @@ trait StoreContext
                 throw new Exception("$thingName does not have a $thingProperty property");
             }
 
-            $string = str_replace($match, $this->getValueForInjection($thingProperty, $thing), $string);
+            $string = str_replace(
+                $match,
+                $this->getValueForInjection($thingProperty, $thing, $thingFormat),
+                $string
+            );
         }
 
         return $string;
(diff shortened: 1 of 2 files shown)


medology-bugman:
Paging @goodevilgenius to review the testing work in this pull request!

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -144,10 +153,13 @@ public function injectStoredValues($string, callable $onGetFn = null, callable $
...
         }
 
-        preg_match_all('/\(the ([^\)]+) of the ([^\)]+)\)/', $string, $matches);
+        preg_match_all('/\(the ([^\)]+) of the ([^\)]+?)( formatted as ([^\)]+))?\)/', $string, $matches);
         foreach ($matches[0] as $i => $match) {
             $thingName = $matches[2][$i];
             $thingProperty = str_replace(' ', '_', strtolower($matches[1][$i]));
+            $thingFormat = $matches[4][$i];
+
+            //dump($matches);
```
goodevilgenius: Please delete this debugging code.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -20,6 +21,14 @@
     /** @var array */
     protected $registry;
 
+    protected static $FORMAT_MYSQL_DATE_AND_TIME = 'a MySQL date and time';
+    protected static $format_map = [
```
johnnie502: We already have a `CarbonDateTime` behat context in STDcheck that used to parse different date, I am wondering if we can add the `12 hour format` to this format collection

- [ ] Done
- [ ] Approve

src/Behat/FlexibleMink/Context/StoreContext.php
johnnie502: Please remove this 

- [ ] Done
- [ ] Approve

Chekote:
@johnnie502 Yes, I will definitely add tests to this. Thanks.

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -254,7 +257,7 @@ protected function formatDateTime(DateTime $dateTime, $thing, $format = null)
         } elseif (is_object($thing)) {
             $value = $this->formatDateTimeFromHostObject($dateTime, $thing);
         } else {
-            $value = (string) $dateTime;
+            return $this->formatDateTimeWithoutHostObject($dateTime);
```
goodevilgenius: If you're going to get rid of assigning `$value` in this block, you might as well do the same in the other two, and remove the final `return` statement.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -277,28 +280,51 @@ protected function formatDateTime(DateTime $dateTime, $thing, $format = null)
...
+     *
+     * @param  DateTime $dateTime the date time to format.
+     * @return string   the result of calling __toString() on the date time, or formatting it as ISO8601 if no
+     *                           __toString method exists.
+     */
+    protected function formatDateTimeWithoutHostObject(DateTime $dateTime)
+    {
+        return method_exists($dateTime, '__toString')
+            ? (string) $dateTime
+            : $dateTime->format(DateTime::ISO8601);
```
goodevilgenius: It would be a good idea to make this a protected property on the class, so that any classes overriding this class can easily change it to their own default format.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -277,28 +280,51 @@ protected function formatDateTime(DateTime $dateTime, $thing, $format = null)
...
+        if (isset($object->$propertyName)) {
+            $value = $object->$propertyName;
+        } else {
+            try {
+                $property = new ReflectionProperty(get_class($object), $propertyName);
+                $property->setAccessible(true);
+
+                $value = $property->getValue($object);
+            } catch (ReflectionException $e) {
+                echo "Caught ReflectionException\n";
```
goodevilgenius: I don't think we want to output anything in this case. That's not typical of a behat step.

- [ ] Done
- [ ] Approved

src/Behat/FlexibleMink/Context/StoreContext.php
Chekote: Yes, I actually meant to do that, but I missed this one.

PR #158, opened 2018-06-27
Implement horizontal scrolling

Chekote:
# Requirement:
I need to test a page where the content causes a horizontal scroll bar to be present. While filling out the form the page is automatically scrolled to the right. I need to be able to scroll back to the left to interact with an element that is now outside of the viewport.

# Implementation:
I modified the scrollWindowToBody function to allow any valid combination of top, bottom, left right so that we can scroll in any direction as necessary.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -647,14 +647,30 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      *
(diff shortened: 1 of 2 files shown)


medology-bugman:

Chekote:
I was wondering how I could implement testing for such a feature, and I couldn't think of a way to do it easily without pulling in my custom fork of the Selenium 2 driver. I didn't think it was a good idea to do that to the FlexibleMink project, as I feel it would cause major composer dependency issues with projects that use it.

PR #170, opened 2018-11-17
Exception cleanup

Chekote:
## Problem:
Exception handling has various issues throughout the project. e.g.:

* Generic exceptions are thrown, making it almost impossible for the caller to handle specific situations.
* Many exceptions are bubbling up without proper `@throws` annotations.

## Fix:
Some fixes remain, but this PR resolves the issues for the Spinner, FlexibleContext, and StoreContext classes by replacing generic exceptions with specific ones, and updating all docblocks with appropriate `@throws` annotations.


medology-bugman:

Chekote:
@johnnie502 yeah... When replace the inject stored value functionality with the BehatNounStoreExtension, it'll help a lot. The extension will inject the stored values with an argument transformer before the context class is even called, so the context class will never be exposed to any exceptions relating to injecting values.

PR #172, opened 2018-11-18
Fix getThingProperty does not build key

Chekote:
Problem:
getThingProperty is accepting an index as an argument and passing it onto Assert::keyExists, but Assert::keyExists does not accept that argument.

Fix:
Used the Key service to build the nth key before passing it on to Assert::keyExists.


medology-bugman:

PR #174, opened 2018-11-18
Add AsyncMink class

Chekote:
# Requirement:
There are various instances where Behat steps are not "Then" steps, but they require a certain level of retry functionality. For example, with complex "When" steps that interact with the page and then need to wait for a button to be visible etc before proceeding. Since currently the only retry functionality is implemented in the BehatRetryExtension, which only works when steps are invoked from the feature file, none of these use cases are catered to.

# Implementation:
I wrote an AsyncMink class which can be used to programmatically make assertions via the Spinner::waitFor mechanism. This prevents having all users of the FlexibleMink library from having to sprinkle waitFor calls throughout their code by centralizing the behavior in this class.

This is purely for use when calling these methods programmatically. This class is NOT a context class and should not be used as such. Async functionality for context classes should be handled by use of the BehatRetryExtension https://github.com/Chekote/BehatRetryExtension

Please note that this class references exception that is either thrown or do not exist yet. This is because this PR is made assuming #170 is merged first.

### DIFF
--- a/src/Medology/Behat/Mink/AsyncMink.php
+++ b/src/Medology/Behat/Mink/AsyncMink.php
@@ -1,0 +1,46 @@
+<?php namespace Medology\Behat\Mink;
+
+use Exception;
+use Medology\Spinner;
+
+/**
+ * Wraps the FlexibleMink context to provide asynchronous assertions.
+ *
+ * The methods here will wait for either the assertion to pass, or for the timeout to expire.
+ *
+ * See Spinner::$default_timeout
+ *
+ * This is purely for use when calling these methods programmatically. This class is NOT a context class and should
+ * not be used as such. Async functionality for context classes should be handled by use of the BehatRetryExtension
+ * https://github.com/Chekote/BehatRetryExtension.
+ *
+ * @mixin FlexibleContext
+ */
+class AsyncMink
+{
+    /** @var FlexibleContext */
+    protected $flexibleContext;
+
+    /**
+     * @param FlexibleContext $flexibleContext the instance to wrap.
+     */
+    public function __construct(FlexibleContext $flexibleContext)
+    {
+        $this->flexibleContext = $flexibleContext;
+    }
+
+    /**
+     * Proxy method to handle all calls and relay them to FlexibleContext wrapped in a waitFor.
+     *
+     * @param  string    $method    the name of the method that was called.
+     * @param  array     $arguments the arguments that were passed to the method.
+     * @throws Exception if the assertion did not pass before the timeout was exceeded.
+     * @return mixed     the result of the lambda if it succeeds.
+     */
+    public function __call($method, array $arguments)
+    {
+        return Spinner::waitFor(function () use ($method, $arguments) {
+            return call_user_func_array([$this->flexibleContext, $method], $arguments);
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


medology-bugman:

johnnie502:
This branch is based on the PR https://github.com/Medology/FlexibleMink/pull/170 that has the same few commits, is this intended?

src/Medology/Behat/Mink/AsyncMink.php
```diff
@@ -0,0 +1,733 @@
...
+     * @throws SpinnerTimeoutException          If the timeout expires before assertion could be attempted even once.
+     * @throws UnsupportedDriverActionException When operation not supported by the driver.
+     * @return NodeElement                      The option.
+     */
+    public function assertVisibleOption($locator)
+    {
+        return Spinner::waitFor(function () use ($locator) {
+            return $this->flexibleContext->assertVisibleOption($locator);
+        });
+    }
```
RedEnchilada: Would it be possible to implement a `__call` method that can spin the entire `FlexibleContext` without repeating all of the docblocks from it and writing redundant code? If the `@mixin` class docblock annotation works as expected, it should keep the IDE happy too.

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -36,12 +41,31 @@ class FlexibleContext extends MinkContext
         'tab'        => 9,
     ];
 
+    /** @var AsyncMink */
+    protected $wait;
+
+    public function __construct()
+    {
+        $this->wait = new AsyncMink($this);
```
RedEnchilada: I don't see a class property for `$this->wait`. Could you add one?

RedEnchilada:
It looked like some of the code was from #170 ? So I kind of skimmed over that part. Hopefully I didn't miss anything important lol

src/Medology/Behat/Mink/AsyncMink.php
Chekote: I considered doing that but then I would not be able to declare the additional `@throws` introduced by using `waitFor`. Do you have a suggestion on how to accomplish that while using `__call`?

Chekote:
The merge of #170 was accidental. I rebased the branch to fix that since it's now merged. Sorry for the rebase after code review >.<

src/Medology/Behat/Mink/FlexibleContext.php
Chekote: It was there. But since I did a force push you'll just have to take my word for it (or call me a liar 😛️).

src/Medology/Behat/Mink/AsyncMink.php
RedEnchilada: Ah, there are extra exceptions. I didn't see them since the docblocks had a ton already and I kind of skimmed through them. Serves me right, I guess

(If `waitFor` used a do-while block and obsoleted the `SpinnerTimeoutException`, that would resolve the additional exception being added. That does go out of the scope of this PR though, so I'm not too concerned.)

Chekote:
I'm going to hold off on merging this for now since if I can get FlexibleMink to use the upcoming BehatNounStoreExtension, it'll significantly reduce the number of exceptions that will bubble up from FlexibleMink.

src/Medology/Behat/Mink/AsyncMink.php
Chekote: Interesting idea. I'll certainly consider that for FlexibleMink 3.0. I wouldn't be able to drop the exception for 2.0 since that'd be a breaking change.

PR #175, opened 2018-11-19
Move assertElementOnPage from STDcheck to FlexibleMink

Chekote:
Various methods were added to STDcheck's WebContext instead of adding them to FlexibleMink. This is the first of many PRs that will move these methods so that all projects using FlexibleMink can utilize them.


healthlabs-bugman:

Issue #169, opened 2018-11-07
I should see ... does not actually assert that the user can see the text in the DOM.


aserv92:
Any behat "I should see ..." assertion does not actually assert that the user can see something in the DOM.

PR #186, opened 2018-11-20
Assert that an element is visible to the user.

aserv92:
bug #169 

Method:
The "I should see assertion"

Expected Behavior:
Asserts that the user can see text on the screen.

Actual Behavior:
Asserts that text is anywhere in the DOM, but not necessarily visible to the user.

Cause:
The "I should see assertion" checks the DOM for text

Solution:
Check the position and visibility of all parent elements of a specified element that contains the text you are looking for.

Example: 

`"Test Text" should be fully displayed`

`<span data-qa-id="Test Text">Test Text</span>`

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -8,12 +8,12 @@ use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Driver\Selenium2Driver;
 use Behat\Mink\Element\NodeElement;
 use Behat\Mink\Element\TraversableElement;
-use Behat\Mink\Exception\ElementNotFoundException;
 use Behat\Mink\Exception\ExpectationException;
 use Behat\Mink\Exception\ResponseTextException;
 use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Behat\MinkExtension\Context\MinkContext;
 use InvalidArgumentException;
+use JMS\Serializer\Tests\Fixtures\Node;
 use ZipArchive;
 
 /**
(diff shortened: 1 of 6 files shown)


healthlabs-bugman:

aserv92:
Manually reassigned reviewers on PR#191 to match this one because both versions needed the same update.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+        $qaId = $this->injectStoredValues($qaId);
+
+        $xpath = '//*[@data-qa-id="' . $qaId . '"]';
+
+        $nodeElement = $this->waitFor(function () use ($xpath) {
+            return $this->getSession()->getPage()->find('xpath', $xpath);
+        });
+
+        if (!$nodeElement instanceof NodeElement && !$not) {
+            throw new ElementNotFoundException(
```
juanmrad: This might not be the best exception if you are doing a message as this exception expects:
`($driver, $type = null, $selector = null, $locator = null)`
I will suggest using:
```
throw new ExpectationException("The node element with qaId $qaId was " . (!$not ? 'not ' : '') . "found ", $this->getSession());
```

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+     * @param $element NodeElement
+     * @param $not boolean asserts NodeElement is partially or not visible in the viewport.
+     * @throws ResponseTextException
+     */
+    public function assertNodeIsFullyVisibleInViewPort(NodeElement $element, $not = false)
+    {
+        $session = $this->getSession();
+
+        if (!$element instanceof NodeElement) {
+            throw new ElementNotFoundException($session, 'Invalid node sent to ' . __FUNCTION__);
```
juanmrad: Same as previous comment this exception might not be the best to throw given the parameters.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+        if (count($parents) < 2) {
+            throw new ResponseTextException('Invalid number of node elements', $session);
+        }
+
+        $elementViewportRectangle = $this->getElementViewportRectangle($element);
+
+        $elementIsVisible = $element->isVisible();
+
+        if (!$not && !$elementIsVisible) {
+            throw new ResponseTextException(
```
juanmrad: This exception should be thrown when an expectation on the response text fails. this should be a `ExpectationException` instead.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+            return;
+        }
+
+        $allAreIn = true;
+
+        foreach ($parents as $parent) {
+            $parentIsVisible = $parent->isVisible();
+
+            if (!$not && !$parentIsVisible) {
+                throw new ResponseTextException(
```
juanmrad: same as previous comment.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+            }
+
+            $parentViewportRectangle = $this->getElementViewportRectangle($parent);
+
+            $isIn = $elementViewportRectangle->isIn($parentViewportRectangle, $not);
+
+            $allAreIn = $allAreIn && !$isIn;
+
+            if (!$not && !$isIn) {
+                throw new ResponseTextException(
```
juanmrad: same as previous comment.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+                    'Node is not fully visible in the viewport.', $session
+                );
+            }
+        }
+
+        if ($not && $allAreIn) {
+            throw new ResponseTextException(
```
juanmrad: same as previous comment

src/Behat/FlexibleMink/Models/Geometry/Rectangle.php
```diff
@@ -0,0 +1,68 @@
...
+class Rectangle
+{
+    public $corner1x = 0;
+    public $corner1y = 0;
+    public $corner3x = 0;
+    public $corner3y = 0;
+
+    /**
+     * Rectangle constructor.
+     *
```
jose-ochoa: Please keep only one space between sections, here and all other areas in this file.

- [ ] Addressed
- [ ] Approved

web/ScrollableDivs.html
```diff
@@ -0,0 +1,307 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <title>All Centered</title>
+        <script>
+          /**
+           *
+           * centers all scrolls with class
```
jose-ochoa: Please capitalize the first letter.

- [ ] Addressed
- [ ] Approved

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
 
         return $aRect['top'] - $bRect['top'];
     }
+
+    /**
+     * Asserts that a qaId is visible and inside the viewport.
+     *
+     * @Then /^"(?P<qaId>[^"]+)" should(?P<not> not|) be fully displayed$/
```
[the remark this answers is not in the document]
aserv92: What do you think about this? I think the assertion `/^The element "(?P<qaId>[^"]+)" should(?P<not> not|) be on my view$/` make sense. What if I change the name of the function to better match the assertion?

    /**
     * Asserts that a qaId is fully displayed inside the viewport.
     *
     * @Then /^"(?P<qaId>[^"]+)" should(?P<not> not|) be fully displayed$/
     *
     * @param string    $qaId   The qaId of the dom element to find
     * @param boolean   $not    asserts qaId is partially or now visible in the viewport.
     * @throws ExpectationException
     */
    public function assertQaIDIsFullyDisplayed($qaId, $not = false)
    {
    ...

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+     *
+     * @param $qaId string The qaId of the dom element to find
+     * @param $not boolean asserts qaId is partially or now visible in the viewport.
```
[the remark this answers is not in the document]
aserv92: It looks like there is a typo. I meant
`* @param $not boolean asserts qaId is partially or not visible in the viewport.`
If we are checking for not fully visible, then the element may still be partially visible or not visible at all.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,170 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+     */
+    public function assertNodeIsFullyVisibleInViewPort(NodeElement $element, $not = false)
+    {
+        $session = $this->getSession();
+
+        if (!$element instanceof NodeElement) {
+            throw new ElementNotFoundException($session, 'Invalid node sent to ' . __FUNCTION__);
+        }
+
+        $driver = $session->getDriver();
```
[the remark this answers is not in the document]
aserv92: $driver is used twice. It's also in the if expression below and used to throw an ExpectationException.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,168 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
...
+                        'One of the node elements parents is not visible', $driver
+                    );
+                }
+
+                return;
+            }
+
+            $parentViewportRectangle = $this->getElementViewportRectangle($parent);
+
+            $isIn = $elementViewportRectangle->isFullyIn($parentViewportRectangle, $not);
```
juanmrad: There is no need to create the variable that will be immediately used. 
```
$isIn = $elementViewportRectangle->isFullyIn($this->getElementViewportRectangle($parent), $not);
$allAreIn = $allAreIn && !$isIn;
```

src/Behat/FlexibleMink/Context/FlexibleContext.php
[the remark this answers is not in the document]
juanmrad: I would prefer visible as we are the using trying to see an element. over it being displayed which could mean it is not there or not loaded properly.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -796,4 +798,168 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
 
         return $aRect['top'] - $bRect['top'];
     }
+
+    /**
+     * Asserts that a qaId is fully visible.
+     *
+     * @Then /^"(?P<qaId>[^"]+)" should(?P<not> not|) be fully visible/
```
juanmrad: missing `$` before the `/`

PR #195, opened 2018-11-21
Move waitForPageLoad from STDcheck to FlexibleMink

Chekote:

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -872,4 +872,20 @@ class FlexibleContext extends MinkContext
 
         return ($aRect['top'] < $bRect['top']) ? -1 : 1;
     }
+
+    /**
+     * Waits for the page to be loaded.
+     *
+     * This does not wait for any particular javascript frameworks to be ready, it only waits for the DOM to be
+     * ready. This is done by waiting for the document.readyState to be "complete".
+     */
+    public function waitForPageLoad($timeout = 120)
+    {
+        $this->waitFor(function () {
+            $readyState = $this->getSession()->evaluateScript('document.readyState');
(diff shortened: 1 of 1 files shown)


healthlabs-bugman:

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -872,4 +872,20 @@ protected function compareElementsByCoords(NodeElement $a, NodeElement $b)
 
         return ($aRect['top'] < $bRect['top']) ? -1 : 1;
     }
+
+    /**
+     * Waits for the page to be loaded.
+     *
+     * This does not wait for any particular javascript frameworks to be ready, it only waits for the DOM to be
+     * ready. This is done by waiting for the document.readyState to be "complete".
+     */
```
awesomeousPrime: Please add the exception being thrown in the docblocks

- [ ] Done
- [ ] Approved

Chekote: The vast majority of methods in FlexibleMink 1.0 are either missing docblocks or have a lot of missing information.

The reason I am moving these methods is to get us off of FlexibleMink 1.0, onto 2.0, and ultimately break the package into smaller more focused and well tested/documented packages such as the BehatRetryExtension and BehatNounStoreExtension.

Considering the state of 1.0's codebase and the fact that we're moving away from it, I didn't see much value in fixing up the docblocks. I just moved then as-is from STDcheck.

I can add them to this methods docblock, but the vast majority of the other methods in this package will not (and won't ever) be to the same standards.

What do you think?

Chekote: I'm going to merge this since it seems you're out of the office today. If you still would like me to add the throws after reading my response, just slack me and I'll take care of it. Thanks.

PR #198, opened 2018-11-28
Creating Method to Assert Element is not Covered by Another

awesomeousPrime:
Adding a method to assert that an element is not being covered by another.  

This method gets the surrounding square space of an item and checks points on the item and compares them to the element being tested.  If there is an item that is found to be different, than the one specified, an exception is thrown.  

A simple test was added to assert that the assertion is working as intended.

Another pull request exists for adding this in 2.0 also [here](https://github.com/Medology/FlexibleMink/pull/199)

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -1189,8 +1189,8 @@ class FlexibleContext extends MinkContext
      * Keep in mind that at the moment, this method performs a check in a square area so this may not work
      * correctly with elements of different shapes.
      *
-     * @param  NodeElement              $element   The element to assert that is not covered by something else.
-     * @param  int                      $leniency  Percent of leniency when performing each pixel check.
+     * @param  NodeElement              $element  The element to assert that is not covered by something else.
+     * @param  int                      $leniency Percent of leniency when performing each pixel check.
      * @throws ExpectationException     If element is found to be covered by another.
      * @throws InvalidArgumentException The threshold provided is outside of the 0-100 range accepted.
      */
@@ -1218,8 +1218,7 @@ JS
         $x = $coordinates['left'];
         $y = $coordinates['top'];
 
-
-        $xSpacing = ($width * ($leniency/100)) ?: 1;
+        $xSpacing = ($width * ($leniency / 100)) ?: 1;
         $ySpacing = ($height * ($leniency / 100)) ?: 1;
 
         $expected = $element->getOuterHtml();
@@ -1236,7 +1235,6 @@ JS
             while ($x < $xLimit) {
                 $found = $this->getSession()->evaluateScript("return document.elementFromPoint($x, $y).outerHTML;");
                 if ($expected != $found) {
-
                     throw new ExpectationException(
                         'An element is above an interacting element.',
                         $this->getSession()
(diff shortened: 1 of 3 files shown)


awesomeousPrime:
!force code review

healthlabs-bugman:
Paging @fortoulrodolfo to review the testing work in this pull request!

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -1101,4 +1101,87 @@ private function getListOfAllNodeElementParents(NodeElement $NodeElement, $stopA
...
         return $NodeElements;
     }
+
+    /**
+     * Step to assert that the specified element is not covered.
+     *
+     * @param  string               $identifier Element Id to find the element used in the assertion.
+     * @throws ExpectationException If element is found to be covered by another.
+     *
+     * @Then the :identifier element should not be covered by another
```
fortoulrodolfo: Minor change. The step should be above the params docblock. Its up to you to address this change. The rest looks good! 👍

web/covering-elements.html
```diff
@@ -0,0 +1,31 @@
...
+
+        #coveringDiv{
+            background-color: firebrick;
+            z-index: 999;
+        }
+    </style>
+</head>
+<body>
+    <div class="samePositionElements" id="testedDiv"></div>
+    <div class="samePositionElements" id="coveringDiv"></div>
```
RedEnchilada: Please add some test cases to cover the following:

- Overlapping elements that aren't in the upper-left corner of the screen failing the assertion
- Partially-overlapping elements (ideally including a check on each corner and in just the center) failing the assertion
- Element that isn't covered by anything passing the assertion
- Element that covers another element passing the assertion that the first element is not covered

awesomeousPrime:
@RedEnchilada Wow very nice catch!

I've gone ahead and made some fixes to the method.  Instead of threshold I've changed it to accept leniency and fixed the formulas and some other issues that came along.  

Here are some screenshots of the results: 
(The semi transparent blue dots or areas, areas are where the script is performing checks on the shape).

0% Leniency: 
screenshot_0

5% Leniency: 
screenshot_5

10% Leniency: 
screenshot_10

25% Leniency: 
screenshot_25

50% Leniency: 
screenshot_50

75% Leniency (Accuracy sucks past 50%): 
screenshot_75

Since the areas being checked are starting from top left and going to bottom right, if there are items with sizes that don't divide nicely, the script will not read the outer edges well, but this can be resolved by decreasing the leniency. 

I'm now adding additional test items requested.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -1167,6 +1167,115 @@ private function getListOfAllNodeElementParents(NodeElement $NodeElement, $stopA
...
+     *
+     * @param int    $x     X coordinate where to place dot.
+     * @param int    $y     Y coordinate where to place dot.
+     * @param string $color The hex color code for the color of the point being drawn.  Keep in mind, the item will
+     *                      appear semi transparent.
+     *
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     * @throws DriverException                  When the operation cannot be done
+     */
+    protected function stipple($x, $y, $color = '#0f0')
```
RedEnchilada: This method is unused. Please remove it unless you have a use for it.

PR #199, opened 2018-11-28
Creating Method to Assert Element is not Covered by Another 

awesomeousPrime:
Adding a method to assert that an element is not being covered by another.

This method gets the surrounding square space of an item and checks points on the item and compares them to the element being tested. If there is an item that is found to be different, than the one specified, an exception is thrown.

A simple test was added to assert that the assertion is working as intended.

Another pull request exists for adding this in 1.0 also [here](https://github.com/Medology/FlexibleMink/pull/198)

### DIFF
--- a/features/FlexibleContext/assertElementIsNotCovered.feature
+++ b/features/FlexibleContext/assertElementIsNotCovered.feature
@@ -1,0 +1,13 @@
+Feature:  Assert Checkbox
+  In order to ensure that elements are interactive
+  As a developer
+  I should have assertions for elements covering others
+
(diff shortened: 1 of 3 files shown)


awesomeousPrime:
!reassign front-end @johnnie502

healthlabs-bugman:

awesomeousPrime:
!reassign back-end @RedEnchilada

healthlabs-bugman:

awesomeousPrime:
!reassign testing @fortoulrodolfo

PR #202, opened 2018-11-28
Move assertJavascriptVariables from STDcheck to FlexibleMink

Chekote:

### DIFF
--- a/src/Behat/FlexibleMink/Context/JavaScriptContext.php
+++ b/src/Behat/FlexibleMink/Context/JavaScriptContext.php
@@ -121,4 +121,21 @@ trait JavaScriptContext
             );
         }
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -121,4 +121,21 @@ public function assertJavascriptVariable($variableName, $expectedValue)
             );
         }
     }
+
+    /**
+     * Asserts that a set of javascript variables have specified values.
+     * The $table should have the variable name in the first column, and the value in the second.
+     *
+     * @Then   the javascript variables should be:
```
jose-ochoa: Please reduce the space after the `@Then`

- [ ] Addressed
- [  ] Approved

PR #210, opened 2018-12-05
Modifying assertElementIsNotCovered method to account for elements in the tested element 2

awesomeousPrime:
At the moment the assertElementIsNotCovered method does not account for items within the element.

If an element has a header for example, the method is at the moment failing because at that point the header is on top of the element specified. This is usually expected to be the case, so I've modified the method to account for this and added a small header to an existing tested element.


healthlabs-bugman:

PR #213, opened 2018-12-11
1.0 refactor

aserv92:
Refactored my changes to flexible mink into small functions without boolean voodoo.

### DIFF
--- a/src/Behat/FlexibleMink/Models/Geometry/Rectangle.php
+++ b/src/Behat/FlexibleMink/Models/Geometry/Rectangle.php
@@ -7,7 +7,6 @@ namespace Behat\FlexibleMink\Models\Geometry;
  */
 class Rectangle
 {
-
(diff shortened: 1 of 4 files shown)


aserv92:
!force code review

healthlabs-bugman:

aserv92:
Reassigned reviewers to match 2.0 refactor that has already been merged.
PR#211

src/Behat/FlexibleMink/Models/Geometry/Rectangle.php
```diff
@@ -7,74 +7,47 @@
  */
 class Rectangle
```
PrinceMegahit: Usually `Rectangle` classes are implemented as an X,Y `Point` tuple and a width,height `Point` tuple, with properties for accessing right left topleft topright etc. But this is great as it stands.

PR #217, opened 2018-12-18
2.0 extended visibility Methods

aserv92:
Add tests to check for element visibility in the viewport and in the document.


healthlabs-bugman:

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+     */
+    public function assertQaIDIsVisibleInViewport($qaId)
+    {
+        $this->flexibleContext->waitForPageLoad();
+
+        $element = $this->getNodeElementByQaID($this->storeContext->injectStoredValues($qaId));
+
+        if (!$element) {
+            throw new ExpectationException(
+                "Data QA ID '$qaId' is not visible, but it should be",
```
Chekote: This should probably be more specific. If you don't have an `$element`, not only is it not visible, but it's not even in the DOM. So something like `Data QA ID '$qaId' was not found in the document` seems more informative.

- [ ] Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+            throw new ExpectationException('Node is not visible in the viewport.',
+                $this->flexibleContext->getSession()->getDriver()
+            );
+        }
+    }
+
+    /**
+     * Asserts that a qaId is not visible in the viewport.
+     *
+     * @Then /^"(?P<qaId>[^"]+)" should not be visible in the viewport$/
```
Chekote: This can also use the non-regex approach.

- [ ] Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+            throw new ExpectationException('Node is visible in the viewport.',
+                $this->flexibleContext->getSession()->getDriver()
+            );
+        }
+    }
+
+    /**
+     * Asserts that a qaId is visible in the document.
+     *
+     * @Then /^"(?P<qaId>[^"]+)" should be visible in the document/
```
Chekote: This can also use the non-regex approach.

- [ ] Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+            throw new ExpectationException('Node is not visible in the document.',
+                $this->flexibleContext->getSession()->getDriver()
+            );
+        }
+    }
+
+    /**
+     * Asserts that a qaId is not visible in the document.
+     *
+     * @Then /^"(?P<qaId>[^"]+)" should not be visible in the document$/
```
Chekote: This can also use the non-regex approach.

- [ ] Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+        $element = $this->getNodeElementByQaID($this->storeContext->injectStoredValues($qaId));
+
+        if (!$element) {
+            return;
+        }
+
+        if ($this->flexibleContext->nodeIsVisibleInDocument($element)) {
+            throw new ExpectationException('Node is visible in the document.',
+                $this->flexibleContext->getSession()->getDriver()
+            );
```
Chekote: Please put all arguments on a separate line when wrapping a method call.

- [ ] Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+     * @Then /^"(?P<qaId>[^"]+)" should not be visible in the document$/
+     *
+     * @param  string                  $qaId
+     * @throws ExpectationException    If the element is fully visible
+     * @throws ReflectionException     If injectStoredValues incorrectly believes one or more closures were
+     *                                      passed. This should never happen. If it does, there is a problem with
+     *                                      the injectStoredValues method.
+     * @throws SpinnerTimeoutException If the timeout expired before the assertion could be run even once.
+     */
+    public function assertQaIDIsNotVisibleInDocument($qaId)
```
Chekote: This is very similar to the `is` variation of the method. We should be able to write a `qaIdIsVisibleInDocument` that returns a true/false instead of throwing an exception, that leverage that in both the `is` and `is not` variations of the assertion methods to remove the duplication.

- [ ] Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+     * @Then /^"(?P<qaId>[^"]+)" should not be visible in the viewport$/
+     *
+     * @param  string                  $qaId
+     * @throws ExpectationException    If the element is fully visible
+     * @throws ReflectionException     If injectStoredValues incorrectly believes one or more closures were
+     *                                      passed. This should never happen. If it does, there is a problem with
+     *                                      the injectStoredValues method.
+     * @throws SpinnerTimeoutException If the timeout expired before the assertion could be run even once.
+     */
+    public function assertQaIDIsNotVisibleInViewport($qaId)
```
Chekote: This is very similar to the `is` variation of the method. We should be able to write a `qaIdIsVisibleInViewport` that returns a true/false instead of throwing an exception, that leverage that in both the `is` and `is not` variations of the assertion methods to remove the duplication.

- [ ] Done

web/ScrollableDivs.html
```diff
@@ -302,5 +303,188 @@
         </div>
     </div>
 </div>
+<button data-qa-id="Button 1">
+    Button 1
+</button>
+<button style="visibility: hidden;" data-qa-id="Button 2">
+    Button 2
+</button>
+<br>
```
juanmrad: might be better to have a div element with a set height over all this `<br>` it is more specific number and we can make sure the elements below are out of the view and also easy to read.

web/ScrollableDivs.html
```diff
@@ -1,51 +1,52 @@
+
```
juanmrad: This is not needed.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -50,4 +50,49 @@ public function isFullyIn(self $rectangle)
...
+            );
+    }
+
+    /**
+     * Checks if this is inside another rectangle.
+     *
+     * @param  Rectangle $rectangle Rectangle to check if this one is inside of
+     * @return bool      returns
+     */
+    public function isIn(self $rectangle)
```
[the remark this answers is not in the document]
ramsey-medology: I think the `assert` prefix should be reserved for functions that throw an exception when an assertion fails. This method simply returns a bool. Something like `$this->contains($rectangle)` reads easier and is better semantically with regard to the rectangle being tested.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -37,17 +39,60 @@ public function __construct($left, $top, $right, $bottom)
...
+            );
+    }
+
+    /**
+     * Checks if this is inside another rectangle.
+     *
+     * @param  Rectangle $rectangle Rectangle to check if this one is inside of
+     * @return bool      returns
+     */
+    public function isIn(Rectangle $rectangle)
```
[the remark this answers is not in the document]
aserv92: Actually I think a better name for for the two intersect methods is simply isInX() and isInY().
What do you think?

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -1397,6 +1397,65 @@ public function nodeIsFullyVisibleInViewport(NodeElement $element)
...
+        $driver = $this->getSession()->getDriver();
+
+        $parents = $this->getListOfAllNodeElementParents($element, 'html');
+
+        if (!$driver->isDisplayed($element->getXpath())) {
+            return false;
+        }
+
+        foreach ($parents as $parent) {
+            if (!$driver->isDisplayed($parent->getXpath())) {
```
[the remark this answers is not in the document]
aserv92: Just now I did, and an unnecessary loop has been removed.

src/Medology/Behat/Mink/QualityAssurance.php
aserv92: We already have a method called nodeIsVisibleInDocument() from FlexibleContext.
These methods in QualityAssurance are finding the NodeElement then returning or throwing specific error messages based what nodeIsVisibleInDocument() returns.

We just refactored this to keep the methods smaller and more readable.

image

src/Medology/Behat/Mink/QualityAssurance.php
aserv92: We already have a method called nodeIsVisibleInViewport() from FlexibleContext.
These methods in QualityAssurance are finding the NodeElement then returning or throwing specific error messages based what nodeIsVisibleInViewport() returns.

We just refactored this to keep the methods smaller and more readable.


screen shot 2018-12-19 at 5 36 44 pm

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -50,4 +50,49 @@ public function isFullyIn(self $rectangle)
...
+     */
+    private function xIntersectsWith(self $rectangle)
+    {
+        return (
+                $this->left >= $rectangle->left &&
+                $this->left <= $rectangle->right
+            ) || (
+                $this->right <= $rectangle->right &&
+                $this->right >= $rectangle->left
+            );
```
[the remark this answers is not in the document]
aserv92: Done. I think.

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -1397,6 +1397,65 @@ public function nodeIsFullyVisibleInViewport(NodeElement $element)
         return true;
     }
 
+    /**
+     * Asserts that a NodeElement is visible in viewport.
+     *
+     * @param  NodeElement                      $element
+     * @throws DriverException
+     * @throws UnsupportedDriverActionException
```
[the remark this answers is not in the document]
aserv92: Done.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -50,4 +50,49 @@ public function isFullyIn(self $rectangle)
...
+     * @return bool      returns
+     */
+    private function yIntersectsWith(self $rectangle)
+    {
+        return (
+                $this->top >= $rectangle->top &&
+                $this->top <= $rectangle->bottom
+            ) || (
+                $this->bottom <= $rectangle->bottom &&
+                $this->bottom >= $rectangle->top
```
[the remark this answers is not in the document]
aserv92: Done. I think.

web/ScrollableDivs.html
aserv92: done

web/ScrollableDivs.html
aserv92: done.

web/ScrollableDivs.html
```diff
@@ -1,51 +1,52 @@
...
-          let scrollWidth = div.scrollWidth;
-          let scrollHeight = div.scrollHeight;
-          div.style.lineHeight = (scrollHeight*1) + 'px';
-          div.scrollLeft = ( scrollWidth  / 2  ) - ( ( clientWidth  - YScrollBarWidth  ) / 2 )  - (YScrollBarWidth/2) + x;
-          div.scrollTop  = ( scrollHeight / 2  ) - ( ( clientHeight - XScrollBarHeight ) / 2 )  - (XScrollBarHeight/2) + y;
+        /**
+         * Centers all scrolls
+         *
+         * @param ClassName
+         * @param x x offset from center
```
juanmrad: no need to have x twice. the name is descriptive. you could on the other hand say: 
`Horizontal offset from center`
`Vertical offset from center`

aserv92: Done.

aserv92: done

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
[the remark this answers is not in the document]
aserv92: Renamed again to overlaps.

src/Medology/Behat/Mink/QualityAssurance.php
aserv92: Re-refactored

src/Medology/Behat/Mink/QualityAssurance.php
aserv92: Re-refactored

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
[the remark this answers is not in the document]
aserv92: Resolved?

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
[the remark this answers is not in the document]
aserv92: Resolved?

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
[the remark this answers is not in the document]
aserv92: Done

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -98,4 +98,126 @@ public function assertQaIDIsNotFullyVisibleInViewport($qaId)
...
+
+        if (!$element) {
+            throw new ExpectationException(
+                "Data QA ID '$qaId' is not visible, but it should be",
+                $this->flexibleContext->getSession()
+            );
+        }
+
+        if (!$this->flexibleContext->nodeIsVisibleInDocument($element)) {
+            throw new ExpectationException('Node is not visible in the document.',
```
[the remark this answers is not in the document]
aserv92: Done.

web/ScrollableDivs.html
aserv92: Okay now it's gone.

src/Medology/Behat/Mink/FlexibleContext.php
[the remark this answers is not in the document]
Chekote: Can you please elaborate. Did you determine that calling `isDisplayed` on an element will take into account the visibility of the elements parents?

If it does, then you should be able to remove this check entirely from the loop and just call the `overlaps` method.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -50,4 +50,49 @@ public function isFullyIn(self $rectangle)
             $this->top >= $rectangle->top &&
             $this->bottom <= $rectangle->bottom;
     }
+
+    /**
+     * Checks if rectangle intersects in x.
+     *
+     * @param  Rectangle $rectangle Rectangle to check if this one is inside of
+     * @return bool      returns
```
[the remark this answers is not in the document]
Chekote: It looks like you removed the entire returns annotation. I just requested that you remove the "returns" description.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -50,4 +50,49 @@ public function isFullyIn(self $rectangle)
...
+                $this->right <= $rectangle->right &&
+                $this->right >= $rectangle->left
+            );
+    }
+
+    /**
+     * Checks if rectangles intersects in y.
+     *
+     * @param  Rectangle $rectangle Rectangle to check if this one is inside of
+     * @return bool      returns
```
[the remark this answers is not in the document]
Chekote: Same thing here with removing the entire `@return` annotation instead of just the description.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -50,4 +50,49 @@ public function isFullyIn(self $rectangle)
...
+                $this->bottom <= $rectangle->bottom &&
+                $this->bottom >= $rectangle->top
+            );
+    }
+
+    /**
+     * Checks if this is inside another rectangle.
+     *
+     * @param  Rectangle $rectangle Rectangle to check if this one is inside of
+     * @return bool      returns
```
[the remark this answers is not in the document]
Chekote: Same thing here with removing the entire `@return` annotation instead of just the description.

src/Medology/Behat/Mink/QualityAssurance.php
Chekote: I was aware of the `nodeIsVisibleInDocument`. I wasn't suggesting that you write a `nodeIsVisibleInDocument`. I was suggesting that you write a `qaIdIsVisibleInDocument`. It's a similar method but would be working from a qaId instead of a node, and separates the two responsibilities of checking or asserting.

src/Medology/Behat/Mink/QualityAssurance.php
Chekote: I was aware of the `nodeIsNotVisibleInDocument`. I wasn't suggesting that you write a `nodeIsNotVisibleInDocument`. I was suggesting that you write a `qaIdIsNotVisibleInDocument`. It's a similar method but would be working from a qaId instead of a node, and separates the two responsibilities of checking or asserting.

src/Medology/Behat/Mink/QualityAssurance.php
```diff
@@ -28,73 +28,62 @@ class QualityAssurance implements Context
...
-     *                                               the injectStoredValues method.
+     *                                          passed. This should never happen. If it does, there is a problem with
+     *                                          the injectStoredValues method.
      * @throws SpinnerTimeoutException          If the timeout expired before the assertion could be run even once.
-     * @throws UnsupportedDriverActionException Exception thrown by drivers when they don't support the requested action.
+     * @throws ExpectationException             If the element is not fully visible
+     * @throws UnsupportedDriverActionException If driver does not support the requested action.
      */
-    public function assertQaIDIsNotFullyVisibleInViewport($qaId)
+    public function assertVisibilityOfQaId($qaId, $place, $visibility = '', $not = false)
```
Chekote: I'm not a fan of how all of these methods were combined into one. When I suggested refactoring, I was suggesting that the shared functionality of multiple methods be refactored into smaller composable methods. Instead, the opposite has been done, where multiple methods have been combined into a single monolithic method with more parameters.

As a result of combining methods, the following problems have been introduced:

* Instead of simplifying the Behat step string, we've had to complicate it
* We've introduced double negatives (!$not)
* We've added poor type safety, where `$not` is both a string and a bool
* We've lost the ability for tooling to auto-complete/warn of document vs viewport

Overall I think that this refactor pushes us in completely the wrong direction.

src/Medology/Behat/Mink/FlexibleContext.php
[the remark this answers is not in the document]
aserv92: Yes, I tested isDisplayed and it is taking all of the parents into account.

src/Medology/Behat/Mink/QualityAssurance.php
aserv92: reverted and refactored a different way.

src/Medology/Behat/Mink/QualityAssurance.php
aserv92: reverted and refactored a different way.

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -1397,6 +1398,54 @@ public function nodeIsFullyVisibleInViewport(NodeElement $element)
...
+    {
+        /** @var Selenium2Driver $driver */
+        $driver = $this->getSession()->getDriver();
+
+        if (!$driver->isDisplayed($element->getXpath())) {
+            return false;
+        }
+
+        return true;
+    }
```
Chekote: This method can be simplified to:

```php
    public function nodeIsVisibleInDocument(NodeElement $element)
    {
        return $this->assertSelenium2Driver('Check if element is displayed')->isDisplayed($element->getXpath());
    }
```

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -37,17 +37,65 @@ public function __construct($left, $top, $right, $bottom)
...
+        $leftOverlap = $this->right <= $rectangle->right && $this->right >= $rectangle->left;
+
+        /** @var bool $leftOverlap If overlaps on the right */
+        $rightOverlap = $this->left >= $rectangle->left && $this->left <= $rectangle->right;
+
+        /** @var bool $leftOverlap If overlaps on the left and right */
+        $leftAndRightOverlap = $this->left <= $rectangle->left && $this->right >= $rectangle->right;
+
+        return $leftOverlap || $rightOverlap || $leftAndRightOverlap;
+    }
```
Chekote: There's a minor inefficiency here. Because you're checking all of the ways that the element could overlap without short-circuiting as soon as one is true, you are paying the cost for checking all variations when that may not be necessary.

You can keep the documented variations but short-circuit as soon as you find a match by combining them as follows:

```php
    private function overlapsInX(self $rectangle)
    {
        return
            // If overlaps on the left
            ($this->right <= $rectangle->right && $this->right >= $rectangle->left) ||
            // If overlaps on the right
            ($this->left >= $rectangle->left && $this->left <= $rectangle->right) ||
            // If overlaps on the left and right
            ($this->left <= $rectangle->left && $this->right >= $rectangle->right);
    }
```

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -37,17 +37,65 @@ public function __construct($left, $top, $right, $bottom)
...
+        $topOverlap = $this->top >= $rectangle->top && $this->top <= $rectangle->bottom;
+
+        /** @var bool $isOverlappingAtBottom If the bottom overlaps */
+        $isOverlappingAtBottom = $this->bottom <= $rectangle->bottom && $this->bottom >= $rectangle->top;
+
+        /** @var bool $isOverlappingAtTopAndBottom If the top and bottom overlaps */
+        $isOverlappingAtTopAndBottom = $this->top <= $rectangle->top && $this->bottom >= $rectangle->bottom;
+
+        return $topOverlap || $isOverlappingAtBottom || $isOverlappingAtTopAndBottom;
+    }
```
Chekote: Same thing here with the short-circuit.

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -1370,25 +1370,26 @@ public function waitForPageLoad()
...
         ) {
             return false;
         }
-
         $elementViewportRectangle = $this->getElementViewportRectangle($element);
-
         foreach ($parents as $parent) {
             if (!$parent->isVisible() ||
-                !$elementViewportRectangle->isFullyIn($this->getElementViewportRectangle($parent))
+                !$elementViewportRectangle->contains($this->getElementViewportRectangle($parent))
```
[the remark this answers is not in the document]
aserv92: Got confused with code review. I changed it to isContainedIn()

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
```diff
@@ -37,17 +37,65 @@ public function __construct($left, $top, $right, $bottom)
     }
 
     /**
-     * Checks if this is inside another rectangle.
+     * Checks if this is fully inside another rectangle.
      *
-     * @param  Rectangle $rectangle Rectangle to check if this one is inside of
-     * @return bool      returns
+     * @param Rectangle $rectangle Rectangle to check against this one
```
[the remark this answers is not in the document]
aserv92: My mistake I was testing different inspection settings and forgot to remove these again.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
aserv92: Done.

src/Medology/Behat/Mink/Models/Geometry/Rectangle.php
aserv92: done

src/Medology/Behat/Mink/FlexibleContext.php
aserv92: Done

PR #219, opened 2019-01-02
[1.0] Fix viewport visibility checks not working on scrolled pages

RedEnchilada:
`Then :element should be fully/partially visible in the viewport` does not work when the page is scrolled, since the `<html>` element's reported viewport bounding rectangle is always sized to screen dimensions and snapped to the top-left of the page, regardless of its size in the element inspector. This issue appears to be unique to the `<html>` tag in particular, so this PR excludes that from the viewport checks.

I'll open a similar PR for the 2.0 branch once this is approved and merged, to avoid having overlapping reviews if both were to be reviewed simultaneously.

### DIFF
--- a/features/FlexibleContext/assertElementIsDisplayed.feature
+++ b/features/FlexibleContext/assertElementIsDisplayed.feature
@@ -28,6 +28,12 @@ Feature: Assert an element is displayed or not displayed
     And "Invisible 2" should not be visible in the viewport
     And "Visible 2" should not be visible in the viewport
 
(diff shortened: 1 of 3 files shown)


healthlabs-bugman:
Paging @fortoulrodolfo to review the testing work in this pull request!

RedEnchilada:
!reassign testing

healthlabs-bugman:

RedEnchilada:
!reassign testing

healthlabs-bugman:

PR #220, opened 2019-01-08
Scroll to elements if the visible element is outside of the viewport

RedEnchilada:
The current steps for interacting with form elements, links, etc. all give errors if the element specified is outside of the viewport, even if the element can be scrolled to. This PR changes that behavior so that, if no elements are visible within the viewport, it will loop over the elements again and try to scroll to each one. This still favors elements within the viewport but allows forms to be filled out properly without having to manually scroll to each and every entry.


healthlabs-bugman:

PrinceMegahit:
!reassign back-end @Chekote

healthlabs-bugman:

PrinceMegahit:
I have no problem reviewing this, but it would be ridiculous for this to get merged without @Chekote approval.

Chekote:
I'm going to mull over this for a bit before I decide on whether to approve it or not.

Chekote:
One concern that comes to mind:

# Scenario
* I am on a page with a button named "OK"
* The "OK" button is way down on the current page
* I interact with something in the viewport that opens a modal
* I try to press the "OK" button in the modal that appears

# Expected Behavior
Behat will spin asserting that the "OK" button is visible in the viewport. It is not initially, so Behat continues to spin until the modal appears, then presses "OK" in the modal when it appears.

# Actual Behavior
Behat asserts that an "OK" button is visible in the viewport. It is not initially, so Behat scrolls down the page and presses the wrong "OK" button.

Chekote:
I can see how this functionality would be very useful in a large number of scenarios. But my biggest concern is that it's very opinionated.

Yes, with the current implementation it can be a hassle because you do have to explicitly state to scroll the element into view. But that also means that you have the choice not to and keep the viewport where it is and wait for the assertion to pass.

With this new feature, as helpful as it is, that choice is taken away.

RedEnchilada:
That scenario's problem is an ambiguity on which "OK" button the user wants to press. In that case the test writer should specify that they want to press the "OK" button within the modal. Otherwise, the problem scenario would still be an issue if the other "OK" button happens to land in the viewport.

I think it'd be better for the test writer to specifically state if the button they want pressed has to be within the viewport, and a step for `When I press the on-screen "OK"` (as opposed to `When I press "OK"`, and similar counterparts for the other steps) would allow that to be declared unambiguously.

The current implementation can lead to steps failing from seemingly unrelated changes if, for instance, a new element is added to the top of the page that shifts a formerly on-screen element off-screen, unless every step that interacts with a form element or link is preceded by scrolling to it. `scroll to this input, then fill this input, then scroll to the next input, then fill that input, then etc` is unnatural-sounding and twice as wordy as it needs to be, and makes the feature files read at too low of a level from a business perspective, which as far as I could tell is the main purpose of using Gherkin instead of writing tests directly in PHP. Most people reading a scenario would assume that the user would scroll the page down to try to find the next form element if it's not on-screen, instead of panicking and immediately giving up, unless the step states that it has to be on-screen.

Chekote:
OK. You've convinced me. Please post a message regarding this to the DevChat channel (so Starfish sees it too) clearly explaining that the test writes need to be specific about what they're interacting with so they don't get a false pass on a test. Thanks.

PR #221, opened 2019-01-21
Fix assertFieldContains() calls undefined method

Chekote:
## Steps:
Invoke FlexibleContext::assertFieldContains with valid arguments.

## Expected Behavior:
The method should assert that the specified field contains the specified value.

## Actual Behavior:
The method throws an error stating that an undefined method "injectStoredValues" was called.

## Cause:
The "injectStoredValues" method exists on the StoreContext class, but "assertFieldContains" is calling it on "this".

## Fix:
Update the "assertFieldContains" to call "injectStoredValues" on "this->storeContext".


healthlabs-bugman:
Paging @juanmrad to review the back-end work in this pull request!

PR #223, opened 2019-01-23
Eliminate SpinnerTimeoutException by using do/while

Chekote:
## Problem:
The SpinnerTimeoutException indicates when the timeout expired before a single attempt of the assertion could be made. This is only necessary because the check to enter the while loop could be false on the first iteration. This exception tends to propagate throughout the code and cause a lot of noise in doc blocks etc.

## Fix:
@RedEnchilada suggested changing the while loop to a do/while, which I did. Thanks, @RedEnchilada! Great idea :)


healthlabs-bugman:

PR #226, opened 2019-01-24
selectOption supports context

Chekote:
## Problem:
The selectOption method assumes that there will only ever be one select on the page with the specified field name. This is not the case if the page contains multiple forms. We need a way to specify which select we want to interact with.

## Fix:
I modified the selectOption method to accept an optional context.


healthlabs-bugman:

PR #227, opened 2019-01-24
Allow registration of Store initialization lambdas

Chekote:
## Problem:
When the StoreContext was a trait in FlexibleMink 1.0, it used a method_exists() check to determine if there was a onStoreInitialized() method, and called it if so. This was used by consumers to register certain nouns that should be available to all tests in the suite, such as "Browser" which would provide the IP Address etc of the client's browser. Since the StoreContext is no longer a trait, this method does not work.

## Fix:
I replaced the method_exists() check with a registerInitialization() method that the consumer can pass a lambda to. The lambda will be called immediately after the store is reset between each test scenario, allowing the store to be seeded or modified as necessary.

### DIFF
--- a/src/Medology/Behat/StoreContext.php
+++ b/src/Medology/Behat/StoreContext.php
@@ -23,6 +23,9 @@ class StoreContext extends Store implements Context
     /** @var Assert */
     protected $assert;
 
+    /** @var callable[] array of functions to call after the registry is cleared to initialize it for use */
+    protected $initializations;
+
     /** @var Key the key service for the noun-store */
     protected $key;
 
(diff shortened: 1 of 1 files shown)


healthlabs-bugman:

src/Medology/Behat/StoreContext.php
```diff
@@ -58,11 +62,21 @@ public function clearRegistry()
     {
         $this->reset();
 
-        if (method_exists($this, 'onStoreInitialized')) {
-            $this->onStoreInitialized();
+        foreach ($this->initializations as $callable) {
+            $callable($this);
```
jose-ochoa: Should $callabe be checked to ensure it is callable?

Chekote: I could do that, but it's a protected array and the only way to add an element to it is to call the `registerInitialization()` method which is typed to `callable` so I feel that it's pretty much protected already. Sure, someone could add code that adds a non-callable to the collection in this class, but I feel like that should be caught at code review since the collection is clearly typed as `callable[]`.

PR #231, opened 2019-02-01
Check element with text in QA 1.0

fortoulrodolfo:
For #7319

Added method in QA trait to check a checkbox within a `QA element`.

### DIFF
--- a/src/Behat/FlexibleMink/Context/QualityAssurance.php
+++ b/src/Behat/FlexibleMink/Context/QualityAssurance.php
@@ -2,6 +2,7 @@
 
 use Behat\Mink\Element\NodeElement;
 use Behat\Mink\Exception\ExpectationException;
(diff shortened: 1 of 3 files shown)


src/Behat/FlexibleMink/Context/QualityAssurance.php
```diff
@@ -151,4 +152,43 @@ public function assertQaIDIsNotVisibleInDocument($qaId)
...
+            );
+
+            if (!$targetElement) {
+                throw new ExpectationException(
+                    "Checkbox with label '$checkbox' was not found within '$qaId'",
+                    $this->getSession()
+                );
+            }
+
+            if ($action === 'check') {
```
jose-ochoa: The `$action` argument can only be two values please throw an exception if it is not valid

PR #240, opened 2019-03-30
Upgraded to PHPUnit 5.7.27.

hopeseekr:
Also set the minimum required PHP version to 5.6.0.


Chekote:
Ideally, no one should be using PHP 5.4 at this point, but this project is compatible with PHP 5.4, so upping the minimum required version is not necessary or appropriate.

PR #246, opened 2019-05-16
 Add Assert Elements Exist & Nth Element Exist Steps

juanmrad:
### Issue:
We need to be able to get all elements on the page or the nth element from a page in order to confirm it exist or further steps to use them. 

### Solution:
Add functions to be able to assert elements exist on the page and to get the nth element on the page.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -254,13 +254,13 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      */
-    public function assertElementsExist($element, $selectorType = 'css')
+    public function assertElementsExist($elementsSelector, $selectorType = 'css')
     {
         $session = $this->getSession();
 
-        return $this->waitFor(function () use ($session, $selectorType, $element) {
-            if (!$allElements = $session->getPage()->findAll($selectorType, $element)) {
-                throw new ExpectationException("No '$element' was not found", $session);
+        return $this->waitFor(function () use ($session, $selectorType, $elementsSelector) {
+            if (!$allElements = $session->getPage()->findAll($selectorType, $elementsSelector)) {
+                throw new ExpectationException("No '$elementsSelector' was not found", $session);
             }
 
             return $allElements;
@@ -270,12 +270,12 @@ class FlexibleContext extends MinkContext
     /**
      * {@inheritdoc}
      */
-    public function assertNthElement($element, $nth, $selectorType = 'css')
+    public function assertNthElement($elementSelector, $nth, $selectorType = 'css')
     {
-        $allElements = $this->assertElementsExist($element, $selectorType);
+        $allElements = $this->assertElementsExist($elementSelector, $selectorType);
 
         if (!isset($allElements[$nth - 1])) {
-            throw new ExpectationException("Element $element $nth was not found", $this->getSession());
+            throw new ExpectationException("Element $elementSelector $nth was not found", $this->getSession());
         }
 
         return $allElements[$nth - 1];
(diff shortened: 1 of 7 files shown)


healthlabs-bugman:
Paging @fortoulrodolfo to review the front-end work in this pull request!

.idea/php.xml

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -251,6 +251,35 @@ public function assertElementNotContainsText($element, $text)
         });
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function assertElementsExist($element, $selectorType = 'css')
+    {
+        return $this->waitFor(function () use ($selectorType, $element) {
+            $session = $this->getSession();
```
jose-ochoa: This may  be able to move outside of the `waitFor` and possibly `getPage()` not sure if it has to re-retrieve the contents or when calling findAll it has a fresh  copy of the dom

tests/Behat/FlexibleMink/Context/FlexibleContext/FlexibleContextTest.php
```diff
@@ -0,0 +1,45 @@
...
+    /** @var DocumentElement|PHPUnit_Framework_MockObject_MockObject */
+    protected $pageMock;
+
+    /** @var FlexibleContext|PHPUnit_Framework_MockObject_MockObject */
+    protected $flexible_context;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setUp()
```
jose-ochoa: This setup appears specific to the needs of the new unit tests, I worry some of these may not be used by other unit tests and will cause unnecessary setup, I would suggest moving it into the new unit test or define a less generic class.

juanmrad: Since this methods are all needed for all the functions this will stay there.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -251,6 +251,36 @@ public function assertElementNotContainsText($element, $text)
         });
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function assertElementsExist($element, $selectorType = 'css')
```
fortoulrodolfo: the function name says `Elements` so the variable name should be `elements` because I think you are trying to assert more than one.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -251,6 +251,36 @@ public function assertElementNotContainsText($element, $text)
...
+    {
+        $session = $this->getSession();
+
+        return $this->waitFor(function () use ($session, $selectorType, $element) {
+            if (!$allElements = $session->getPage()->findAll($selectorType, $element)) {
```
fortoulrodolfo: why you are not using the method `assertElementOnPage` or `elementsCount` instead???

src/Behat/FlexibleMink/PseudoInterface/FlexibleContextInterface.php
```diff
@@ -102,6 +102,26 @@ trait FlexibleContextInterface
...
+     */
+    abstract public function assertElementsExist($element, $selectorType = 'css');
+
+    /**
+     * Checks that the nth element exists and returns it.
+     *
+     * @param  string               $element      The elements to search from.
+     * @param  int                  $nth          This is the nth amount of the element.
+     * @param  string|array         $selectorType selector type locator.
+     * @throws ExpectationException
```
fortoulrodolfo: Please add a description when the exception may occur.

src/Behat/FlexibleMink/Context/FlexibleContext.php
juanmrad: I'm returning elements. the variable name there is for the element selector. I'll change the name for it to make it clearer.

src/Behat/FlexibleMink/Context/FlexibleContext.php
juanmrad: Because we want to get all elements on the page and make sure they do exist on the page. Element on page only returns the first element and the count will only return the count of elements.

PR #248, opened 2019-06-04
Moving NodeElement::press into waitFor.

awesomeousPrime:
The pressing of the node element when pressing a button should also be
surrounded by the waitFor method.

There is a scenario that is attempting to scroll to the bottom where a button is and then press it.  What's happening is that at times the scroll is triggered, and the button is attempted to be pressed before the browser has gotten to the desired scroll area. This causes an exception to be thrown but since the `$button->press()` is outside the waitFor lambda, the button pressing is not re-tried.

The pressing of the button should also be included in a `waitFor` to try and recover from an error when this happens.

Issue: Medology/stdcheck.com#7989

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -928,6 +928,8 @@ class FlexibleContext extends MinkContext
 
     /**
      * {@inheritdoc}
+     *
+     * @throws ExpectationException If Button is found but not visible in the viewport.
      */
     public function pressButton($locator)
     {
@@ -940,9 +942,10 @@ class FlexibleContext extends MinkContext
             if ($button->getAttribute('disabled') === 'disabled') {
                 throw new ExpectationException("Unable to press disabled button '$locator'.", $this->getSession());
             }
-
-            $button->press();
         });
+
+        $this->assertNodeElementVisibleInViewport($button);
+        $button->press();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


healthlabs-bugman:

awesomeousPrime:
!reassign testing @Chekote

healthlabs-bugman:

awesomeousPrime:
!reassign back-end @juanmrad

healthlabs-bugman:
Paging @juanmrad to review the back-end work in this pull request!

awesomeousPrime:
Reassigning reviewers to match the changes for the 1.0 branch pull request since that has already been reviewed and the changes are the same.

awesomeousPrime:
Removing reviewers for the moment since changes still have to be made.

awesomeousPrime:
@Chekote 1.0 Branch pull request version of: 
https://github.com/Medology/FlexibleMink/pull/249

I've gone ahead and applied the same code review change items mentioned in the other PR.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -928,6 +928,8 @@ public function hitKey($key)
 
     /**
      * {@inheritdoc}
+     *
+     * @throws ExpectationException If Button is found but not visible in the viewport.
```
Chekote: When a class is implementing an interface (or a pseudo-interface such as with the traits here), you should always put the updated throws on the interface. If the implementation throws an exception that is not declared on the interface, then your implementation violates the interface's contract. There is no way that a developer coding against the interface would know that this exception is a possibility.

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -1241,6 +1244,26 @@ public function waitForPageLoad($timeout = 120)
...
     }
 
+    /**
+     * Asserts that the node element is visible in the viewport.
+     *
+     * @param  NodeElement          $element Element expected to be visble in the viewport.
+     * @throws ExpectationException If the element was not found visible in the viewport.
+     * @throws Exception            If the assertion did not pass before the timeout was exceeded.
+     */
+    public function assertNodeElementVisibleInViewport(NodeElement $element)
```
Chekote: Please add this to FlexibleContextInterface and move the docblock there, replacing the docblock here with `{@inheritdoc}`.

PR #254, opened 2019-07-18
Nested get feature

rashidsaf:
## Feature  
 - Modification in `StoreContext::get()`  method will allow you access chained(nested) object or arrays such as: 

     `$obj1->$obj2->...$objn->attribute` or `$arr1[$arr2['atribute']]`

     Syntax: `Foo's Bar's attribute` or `Foo.Bar.attribute`
- Created `StoreContext::setThingProperty()` method to be able to assign elements to objects/arrays such as:

     `$bazObject->fooProperty = $barObject`

     Syntax: `"barObject" is stored as property "fooProperty" of "bazObject"`

#### Examples:
- Step `I update the "Customer's Order's feature"`
- Step `the following "App\Model\Foo's Bar's Obj" should exist in the system`

### Backend
- Modified `StoreContext::get()` method 
- Created `StoreContext::setThingProperty()` method 

### Testing
- Testing nested object setting and retrieving, also checking method handling on non-object types


healthlabs-bugman:
Paging @smijo149 to review the back-end work in this pull request!

smijo149:
@rashidsaf I couldn't figure out the relation between this PR and the issue attached to it. Please verify if the issue attached to it is correct. And if it is correct then I would like to discuss it with you.

rashidsaf:
> @rashidsaf I couldn't figure out the relation between this PR and the issue attached to it. Please verify if the issue attached to it is correct. And if it is correct then I would like to discuss it with you.

The issue and PR is not directly related but Donald suggested to create this PR. The idea is  to create more generic functionality that I also used for the testing of this Bug issue

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -84,6 +88,68 @@ public function assertIsStored($key, $nth = null)
...
+    /**
+     * Retrieves a value from a nested array or object using array list.
+     * (Modified version of data_get() laravel > 5.6).
+     *
+     * @param  mixed    $target    The target element
+     * @param  string[] $key_parts Key string dot notation
+     * @param  mixed    $default   If value doesn't exists
+     * @return mixed
+     */
+    public function data_get($target, array $key_parts, $default = null)
```
smijo149: In this function I see that the return statement repeats several times, `return $this->closureValue($default);`. So I just came up with an equivalent function. Please verify whether this works,

```php
    public function data_get($target, array $key_parts, $default = null)
    {
        if (empty($key_parts)) {
            return $target;
        }
        foreach ($key_parts as $segment) {
            if ((is_array($target) || $target instanceof ArrayAccess || is_object($target)) && (isset($target[$segment]) || isset($target->{$segment}))) {
                $target = isset($target[$segment]) ? $target[$segment] : $target->{$segment};
            } else {
                return $this->closureValue($default);
            }
        }
        return $target;
    }
```

rashidsaf: I tried, it didn't work. The problem is conditions are tight to each other so mixing all would create a problem. Maybe I should keep as it is , because its also easy to read and understand

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -84,6 +88,68 @@ public function assertIsStored($key, $nth = null)
...
+                if (!isset($target->{$segment})) {
+                    return $this->closureValue($default);
+                }
+                $target = $target->{$segment};
+            } else {
+                return $this->closureValue($default);
+            }
+        }
+
+        return $target;
```
Chekote: This method is a little long. We should extract the body of the 1st three if / else statements into their own methods:

```
getArrayValue()
getArrayAccessValue()
getObjectValue()
```

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -111,11 +177,14 @@ public function get($key, $nth = null)
...
-        if (!$this->isStored($key, $nth)) {
+        list($target_key, $key_parts) = $this->parseKeyNested($key);
+
+        if (!$this->isStored($target_key, $nth)) {
             return;
         }
 
-        return $nth ? $this->registry[$key][$nth - 1] : end($this->registry[$key]);
+        return $nth ? $this->data_get($this->registry[$target_key][$nth - 1], $key_parts) :
+            $this->data_get(end($this->registry[$target_key]), $key_parts);
```
Chekote: When wrapping a ternary onto multiple lines, please place the "body" of the if and else on their own lines:

```php
        return $nth
            ? $this->data_get($this->registry[$target_key][$nth - 1], $key_parts)
            : $this->data_get(end($this->registry[$target_key]), $key_parts);
```

This makes it easier to read because each component of the ternary is neatly isolated on its own line, so the reader can quickly parse either the individual components or the entire ternary as necessary.

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -380,4 +449,38 @@ public function assertThingPropertyContains($thing, $property, $expected)
             throw new Exception("Expected the '$property' of the '$thing' to contain '$expected', but found '$actual' instead");
         }
     }
+
+    /**
+     * Assign the element of given key to the target object/array under given attribute/key.
+     *
+     * @Given /^"([^"]*)" is attached to "([^"]*)" with "([^"]*)" attribute$/
```
Chekote: What's the use-case for this? I have concerns about the assumption that we should immediately call save on an Eloquent model. Also having FlexibleMink have to know about Eloquent is something I'd prefer to avoid if possible.

rashidsaf: As you mentioned above the main purpose is to set things as object property or array value. I couldn't find the method for this purpose So I created one. I used this to test the `get()` method and also it can be used for other purposes. Maybe I should change the given syntax to something like `@Given /^"([^"]*)" is stored as (key|property) "([^"]*)" of  "([^"]*)"$/`

rashidsaf: @Chekote I will remove `save()` method. If needed we can make another test method and wrap this function and gain ability to store eloquent models.

src/Behat/FlexibleMink/Context/StoreContext.php
rashidsaf: Done

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -59,7 +62,8 @@ public function assertThingIs($key, $expected = null)
     {
         if (($actual = $this->get($key)) !== $expected) {
             throw new Exception(
-                "Expected $key to be " . var_export($expected, true) . ', but it was ' . var_export($actual, true)
+                "Expected $key to be " . var_export($expected, true) . ', but it was ' .
+                var_export($actual, true)
```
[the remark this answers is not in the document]
rashidsaf: I reverted it. It should be same as in 1.0 branch

src/Behat/FlexibleMink/Context/StoreContext.php
rashidsaf: Done

Chekote: My apologies, I seem to have lead you astray with this recommendation. I didn't notice the requirement of having the bodies of the if and else have to also return from the function instead of just assigning a value to `$target`.

Now that the change has been made, I can see that although the result is fewer lines in the original method, it's actually harder to read. Additionally, it has introduced a bug where if any of the values are falsey, the default will be returned when it should not have been.

Please revert this commit and the following related style commits and we will go with your original implementation.

rashidsaf: I reverted back to the initial one. Please check

Chekote:
This looks good. Now that we are happy with the implementation, can you please also open a PR to implement this in FlexibleMink 2.0.

PR #256, opened 2019-07-31
Fix the parallelization for PHP 7.1

manuerumx:
### Context
Behat uses the Symfony dependency injection component. So, every single decorator has a priority.

In the case of PHP 7. Behat, fails to load the class on time and is not executed the expected method. 

### Fix
By providing a value for the priority tag attribute you can influence the order in which middlewares/commands are added and loaded.

#### Thanks
@johnnie502


healthlabs-bugman:
Paging @BKorenek to review the back-end work in this pull request!

PR #249, opened 2019-06-04
Moving NodeElement::press into waitFor.

awesomeousPrime:
The pressing of the node element when pressing a button should also be
surrounded by the waitFor method.

There is a scenario that is attempting to scroll to the bottom where a button is and then press it. What's happening is that at times the scroll is triggered, and the button is attempted to be pressed before the browser has gotten to the desired scroll area. This causes an exception to be thrown but since the $button->press() is outside the waitFor lambda, the button pressing is not re-tried.

The pressing of the button should also be included in a waitFor to try and recover from an error when this happens.

Issue: Medology/stdcheck.com#7989

PR #257, opened 2019-07-31
Press button update1.0

fortoulrodolfo:
Merging again changes addressed by Joel on PR https://github.com/Medology/FlexibleMink/pull/249

The pressing of the node element when pressing a button should also be
surrounded by the waitFor method.

There is a scenario that is attempting to scroll to the bottom where a button is and then press it.  What's happening is that at times the scroll is triggered, and the button is attempted to be pressed before the browser has gotten to the desired scroll area. This causes an exception to be thrown but since the `$button->press()` is outside the waitFor lambda, the button pressing is not re-tried.

The pressing of the button should also be included in a `waitFor` to try and recover from an error when this happens.

Issue: Medology/stdcheck.com#7989

### DIFF
--- a/tests/Behat/FlexibleMink/Context/FlexibleContext/PressButtonTest.php
+++ b/tests/Behat/FlexibleMink/Context/FlexibleContext/PressButtonTest.php
@@ -35,6 +35,9 @@ class PressButtonTest extends FlexibleContextTest
         return $flexible_context;
     }
 
+    /**
+     * Asserts an exception is thrown and the press method is not triggered if the node is not visible in the viewport.
+     */
     public function testFailingToSeeNodeElementIsVisibleInViewportPreventsButtonFromBeingPressed()
     {
         // Need mock with original constructor.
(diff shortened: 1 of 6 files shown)


healthlabs-bugman:
Paging @BKorenek to review the testing work in this pull request!

tests/Behat/FlexibleMink/Context/FlexibleContext/PressButtonTest.php
```diff
@@ -0,0 +1,163 @@
...
+                array_merge(['scrollToButton', 'assertNodeElementVisibleInViewport', 'getSession'], $additional_methods)
+            )
+            ->getMock();
+
+        $flexible_context->method('getSession')->willReturn($sessionMock);
+
+        return $flexible_context;
+    }
+
+    public function testFailingToSeeNodeElementIsVisibleInViewportPreventsButtonFromBeingPressed()
```
johnnie502: Please add a description for this test

tests/Behat/FlexibleMink/Context/FlexibleContext/PressButtonTest.php
```diff
@@ -0,0 +1,163 @@
...
+        $this->setExpectedException(get_class($exception), $exception->getMessage());
+
+        $flexible_context->method('assertNodeElementVisibleInViewport')
+            ->willThrowException($exception);
+        $button->expects($this->never())->method('press');
+
+        $flexible_context->pressButton('this is a test');
+    }
+
+    public function testAttemptingToPressDisabledButtonThrowsException()
```
johnnie502: Please add a description for this test

PR #261, opened 2019-08-01
Complex key patch

rashidsaf:
**Issue:** 
Existing Behat tests conflicts with new complex_key feature. There are many key names containing `'s` (example: `Order's account`). 

**Solution:** 
Made the method checking all possible syntax of key. Returning the stored value if key exists as non-parsed version, if not checking for nested key

### DIFF
--- a/features/StoreContext.feature
+++ b/features/StoreContext.feature
@@ -82,7 +82,7 @@ Feature: Store Context
      And the following is stored as "ChildDataObject":
       | attribute  | foo |
      And "ChildDataObject" is stored as property "childData" of "DataObject"
-    Then the "DataObject's childData's attribute" should be "foo"
+    Then the "DataObject`s childData`s attribute" should be "foo"
 
   Scenario: Chained objects/arrays 3rd level retrieves successful
     Given the following is stored as "DataObject":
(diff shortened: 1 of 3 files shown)


healthlabs-bugman:
Paging @manuerumx to review the back-end work in this pull request!
Paging @smijo149 to review the testing work in this pull request!

manuerumx:
Also, all the tests who stores/retrieves a file are failing in [stdcheck](https://circleci.com/gh/Medology/stdcheck.com/51173?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Even with this patch.

`behat features/web/manage/patients/messages/viewAttachment.feature`

rashidsaf:
> Looks like a wrong issue is attached to this PR. Could you please check.

@smijo149 Yes it is not directly releated. But this blocking my main issue thats why I attached.

features/StoreContext.feature
```diff
@@ -82,7 +82,7 @@ Feature: Store Context
      And the following is stored as "ChildDataObject":
       | attribute  | foo |
      And "ChildDataObject" is stored as property "childData" of "DataObject"
-    Then the "DataObject's childData's attribute" should be "foo"
+    Then the "DataObject`s childData`s attribute" should be "foo"
```
Chekote: I don't like this approach to resolving the conflict. It seems arbitrary and unintuitive to use a non-standard character to differentiate between the two. Can we not have a priority approach whereby we check the store for the key "as-is" before we attempt to find an item using the `testParseKeyNested()` method?

johnnie502:
Agree with other reviews. I will wait until the next change to review.

features/StoreContext.feature
rashidsaf: My first implementation was like as you said, but I though I could cause another unexpected tests fail. I reverted back. Will run all the tests on stdcheck and make sure that all passes, than will push the changes

rashidsaf:
I have changed the implementation as suggested. Updated the solution above

PR #263, opened 2019-08-02
Adding TraversableElement as parameter in pressButton and ScrollToBut…

DevanTarunKumar:
- This changes were made to support one of my requirements in another PR where I making sure that the button I'm clicking on belongs to Element ( dialogue box, etc) that I'm intended to click. So, with these changes, we can now be sure and confident that the button we click is the one which we wanted to.   
- As it is by default null, it wont effect other test scenarios.


healthlabs-bugman:

PR #268, opened 2019-08-13
Adding TraversableElement as parameter in pressButton and ScrollToButton

DevanTarunKumar:
updating this changes to v2.0 because these changes are  already updated in v1.0 

## Problem
- we are not sure whether we are clicking the button we intended to. 

## Cause
- pressButton function finds the button with name provided and presses it, without knowing whether we are clicking the right button. 

## Fix
- Fix is to introduce a new parameter TraversableElement, (which will have reference to the element i.e., modal/alert etc) to pressButton and scrolToButton functions. We can now be sure that button we are pressing is right button.  
- As the parameter is by default null, it wont effect other test scenarios.


healthlabs-bugman:
Paging @manuerumx to review the testing work in this pull request!

PR #269, opened 2019-08-16
[V1.0] Assertion is added to check expected option is selected from dropdown

smijo149:
This assertion is need to write a Behat test for a scenario in PR https://github.com/Medology/www.healthlabs.com/pull/1026

This new assertion statement asserts whether the expected option in selected in a dropdown select menu.

### DIFF
--- a/src/Behat/FlexibleMink/Context/FlexibleContext.php
+++ b/src/Behat/FlexibleMink/Context/FlexibleContext.php
@@ -763,6 +763,41 @@ class FlexibleContext extends MinkContext
     }
 
     /**
+     * {@inheritdoc}
+     *
+     * @Then the :field drop down should have the :option selected
+     */
+    public function assertSelectOptionSelected($field, $option)
+    {
+        /** @var NodeElement $selectField */
+        $selectField = $this->waitFor(function () use ($field) {
+            return $this->assertFieldExists($field);
+        });
+
+        $option = $this->injectStoredValues($option);
+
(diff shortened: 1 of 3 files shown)


healthlabs-bugman:
Paging @BKorenek to review the back-end work in this pull request!

src/Behat/FlexibleMink/Context/FlexibleContext.php
```diff
@@ -762,6 +762,41 @@ public function assertSelectContainsExactOptions($select, TableNode $tableNode)
         });
     }
 
+    /**
+     * {@inheritdoc}
+     *
+     * @Then the :field drop down should have the :option selected
```
BKorenek: This should be phrased in accordance with the HTML tags: 

```@Then the :field select should have the :option option selected```

smijo149: Yeah I agree, but the reason why I wrote it that way, is because it is an existing step in STDCheck and I am just trying to focus only on moving it to FlexibleMink so that I can use it for writing tests for HealthLabs. If I make this change, I have to update all the usages on STDCheck as well on another PR. So what do you think?

jose-ochoa: The step definition should include language that the end-user understand, in this case, the end-user doesn't know HTML terms so they would refer to it as `drop down `

PR #287, opened 2019-08-30
Update PullApprove configuration 

aserv92:
Rename the PullApprove configuration. Modify PullApprove conditions and add unmet status explanations.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,11 +1,36 @@
 version: 3
 extends: "https://api.github.com/repos/medology/Scripts/contents/Pull%20Approve/base.pullapprove.yml?ref=master"
 pullapprove_conditions:
-  - "state == 'open'"
-  - "'[DO NOT REVIEW]' not in title"
-  - "base.ref == '1.0'"
-  - "'ci/circleci: build' in statuses.succeeded or 'Force Review' in labels"
-  - "'ci/circleci: build-php5' in statuses.succeeded or 'Force Review' in labels"
-  - "'ci/circleci: test' in statuses.succeeded or 'Force Review' in labels"
-  - "'ci/circleci: test-php5' in statuses.succeeded or 'Force Review' in labels"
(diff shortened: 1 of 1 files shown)


.pullapprove.yml
```diff
@@ -1,11 +1,36 @@
...
+    unmet_status: pending
+    explanation: "Tests and builds must pass before requesting review"
+  - 
+    condition: "'ci/circleci: test-php5' in statuses.succeeded or 'Force Review' in labels"
+    unmet_status: pending
+    explanation: "Tests and builds must pass before requesting review"
+  - 
+    condition: "'continuous-integration/styleci/pr' in statuses.succeeded or 'Force Review' in labels"
+    unmet_status: pending
+    explanation: "StyleCI must pass before requesting review"
```
Chekote: If we're calling out StyleCI specifically in this explanation, I feel like we should be consistent and call out CircleCI and the Job in the other condition's explanations also.

Chekote: I also have a slight concern with how this is implemented. Bugman checks for a "pass" status and will not assign reviewers until that is present. This is checking for the absence of "pending" status. Bugman's approach is more secure because if CircleCI has not responded to the webhook yet by the time it checks, it still waits for the job to run and pass, whereas if PullApprove checks before CircleCI has responded, it will see that there are no pending checks, and will assign reviewers prematurely.

aserv92: Running a test to see if pull waits for success or waits for not pending.

aserv92: To test your concern I changed php5-test and php5-build to php6-test and php6-build in the PullApprove configuration file as we don't build or test this application on PHP6. PHP 6 testing and building is never in the status checks.

The screenshot below confims that pull approve is waiting for php6-test and php6-build to pass before it would assign reviewers.

aserv92: Now the CircleCi tests are explicitly called out.

aserv92: Added CircleCI to explanations

aserv92:
Reassigned reviewer as @manuerumx  is out for the week.

PR #289, opened 2019-09-04
Issue#1035 - Assert Canonical Tag

juanmrad:
Add the same function to 2.0 flexible mink.

Add test to assert canonical tag exist and contains the proper link.
add appropriate needed function to assert any node contain attribute values.

### DIFF
--- a/src/Medology/Behat/Mink/FlexibleContext.php
+++ b/src/Medology/Behat/Mink/FlexibleContext.php
@@ -1273,6 +1273,37 @@ class FlexibleContext extends MinkContext
     }
 
     /**
+     * Asserts that all nodes have the specified attribute value.
+     *
+     * @param  string                           $locator     The attribute locator of the node element.
+     * @param  array                            $attributes  A key value paid of the attribute and value the nodes
+     *                                                       should contain
+     * @param  string                           $selector    The selector to use to find the node.
+     * @param  null                             $occurrences The number of time the node element should be found.
+     * @throws DriverException                  When the operation cannot be done
+     * @throws ExpectationException             If the nodes attributes do not match
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     */
+    public function assertNodesHaveAttributeValues($locator, array $attributes, $selector = 'named', $occurrences = null)
+    {
+        /** @var NodeElement[] $links */
+        $nodes = $this->getSession()->getPage()->findAll($selector, $locator);
+        if (!count($nodes)) {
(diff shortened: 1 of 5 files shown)


src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -1272,6 +1272,37 @@ public function pressButton($locator, TraversableElement $context = null)
...
+     * @throws ExpectationException             If the nodes attributes do not match
+     * @throws UnsupportedDriverActionException When operation not supported by the driver
+     */
+    public function assertNodesHaveAttributeValues($locator, array $attributes, $selector = 'named', $occurrences = null)
+    {
+        /** @var NodeElement[] $links */
+        $nodes = $this->getSession()->getPage()->findAll($selector, $locator);
+        if (!count($nodes)) {
+            throw new ExpectationException("No node elements were found on the page using '$locator'", $this->getSession());
+        } elseif ($occurrences && count($nodes) != $occurrences) {
```
BKorenek: without an else this should probably just be two ifs.

BKorenek: Also ```!==``` instead of !=

web/assert-link-location.html
```diff
@@ -0,0 +1,14 @@
...
+<html>
+
+<head>
+    <title>Text in Element for Flexible Mink</title>
+    <link rel="canonical" href="http://localhost.local/testing" />
+</head>
+
+<body>
+<div id="emptyDiv"></div>
+<div id="divWithText">This is a div</div>
```
BKorenek: tab before each div.

src/Medology/Behat/Mink/LinkContext.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Medology\Behat\Mink;
```
BKorenek: this should be up top right?

web/assert-link-location.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>Text in Element for Flexible Mink</title>
+    <link rel="canonical" href="http://localhost.local/testing" />
+</head>
+
+<body>
+<div id="emptyDiv"></div>
```
BKorenek: standard id convention is with dashes

juanmrad: Using an already existing class :trollface:

PR #295, opened 2019-10-29
Add javascript equals in 2.0

juanmrad:
This function was added on 1.0 but it is missing on 2.0:

https://github.com/Medology/FlexibleMink/pull/292/files


juanmrad:
Closing since there is a very similar function already existing 🤷‍♂

PR #301, opened 2019-11-15
Script to reject 'wait for n seconds' steps in feature files

izazueta:
For https://github.com/medology/team-standards-and-procedures/issues/116

Based on PR https://github.com/medology/stdcheck.com/issues/9042
- Shell script to check and fail if there are 'wait for n seconds' steps on feature files.
- Run check in Circle CI at the end of the test job.

**CLOSE NOTES:** 
- After a discussion on how to share this check across projects, we decided to close this PR for now in favor of a new modular and shared approach for this and other tools. 
- PR https://github.com/medology/www.healthlabs.com/issues/1173 will be deployed to test and gather more feedback about this check.

### DIFF
--- a/bin/reject-wait-seconds.sh
+++ b/bin/reject-wait-seconds.sh
@@ -3,9 +3,8 @@
 set -eu
 
 # colors
-export RED='\033[0;31m'
-export GREEN='\033[0;32m'
-export NC='\033[0m'
+RED='\033[0;31m'
+NC='\033[0m'
 
 # grep flavor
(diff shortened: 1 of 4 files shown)


bin/reject-wait-seconds.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# colors
+export RED='\033[0;31m'
```
jose-ochoa: These seem to be only used in this script the `export` part can be removed

PR #306, opened 2019-11-26
Wrap assertJavascriptVariableType in waitFor

BKorenek:
## Problem
assertJavascriptVariableType fails immediately if the variable is not of the expected type. Since the variable is often set asynchronously, there is a race condition where the assertion fails before it is set.

## Fix
Wrap the method in a waitFor so the assertion can retry until it passes or the timeout expires.


src/Behat/FlexibleMink/Context/JavaScriptContext.php
```diff
@@ -43,16 +43,20 @@ public function assertJavascriptVariableHasAValue($variable)
...
+        $session = $this->getSession();
 
-        // If it doesn't match - we failed.
-        if ($result != $type xor $not) {
-            throw new ExpectationException(
-                "The variable \"$variable\" should$not be type $type, but is $result",
-                $this->getSession()
-            );
-        }
+        $this->waitFor(static function () use ($type, $not, $variable, $session) {
```
Chekote: I never thought of using static for closures, but it makes a lot of sense. It would probably help with beating the shit out of the GC during Behat tests. 🤔

Issue #321, opened 2020-03-11
Image loaded and image check dynamic support


rashidsaf:
### Problem description
We started to use dynamically loading images with no `src` or a placeholder URL instead of the actual image address. But in FlexibleMink we are checking the `src` attribute for comparison

### Acceptance criteria
- [ ] Compare the `currentSrc` DOM property instead of the html `src` attribute of the img
- [ ] Write test

rashidsaf:
Scored with @fortoulrodolfo and @manuerumx

juliohl:
### QA - Local
Spoke with @rashidsaf and the modifications are related to behat tests only. I tried to execute a sanity test round, however, there is no UI interface within the PR so I'm unable to do it. Moving to **Deploy & Test**.

fiyyazp:
will more to review today

fiyyazp:
will move to QA today

BKorenek:
#327 Cannot test flexible locally, only affects testing. Moving to deploy and test. #9999

BKorenek:
No production QA to do, moving to in progress to create PR's against corresponding repo's. #9999

BKorenek:
Moving to approval, I thought more PR's were needed but they are not.

PR #324, opened 2020-03-11
Check image

rashidsaf:
For https://github.com/Medology/FlexibleMink/issues/321

### Problem description
We started to use dynamically loading images with no `src` or a placeholder URL instead of the actual image address. But in FlexibleMink we are checking the `src` attribute for comparison

### Solution
- Refactored the code so now it can test the actual loaded URL (`currentSrc`) instead of the attribute
- Added new test scenario for check the behavior
- Fixed some scripts used in CI build

### DIFF
--- a/bin/init/wait_for_services.sh
+++ b/bin/init/wait_for_services.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -eu
+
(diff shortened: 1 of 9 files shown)


rashidsaf:
> If you haven't already done so, please also submit a PR for 2.0.

I am gonna do it today, was waiting for the code review

PR #327, opened 2020-03-16
Dynamic images assertion support for 2.0

rashidsaf:
For #321

### Problem description
We started to use dynamically loading images with no src or a placeholder URL instead of the actual image address. But in FlexibleMink we are checking the src attribute for comparison

### Solution
Refactored the code so now it can test the actual loaded URL (currentSrc) instead of the attribute
Added new test scenario for check the behavior
Fixed some scripts used in CI build

### DIFF
--- a/features/WebDownloadContext.feature
+++ b/features/WebDownloadContext.feature
@@ -18,4 +18,6 @@ Feature: Web Download Context
 
   Scenario: Developer Can Test if an Image without src loaded dynamically
     When I reload the page
+    Then I should see "/img/placeholder.svg" image in "dynamic2-image"
+    When I press "Load image"
     Then I should see "/img/medology.png" image in "dynamic2-image"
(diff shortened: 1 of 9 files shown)


features/WebDownloadContext.feature
```diff
@@ -7,7 +7,15 @@ Feature: Web Download Context
...
   Scenario: Developer Can Test if an Image Link is Invalid (Broken)
     Then I should not see an image in "invalid-image"
+
+  Scenario: Developer Can Test if an Image loaded dynamically
+    When I reload the page
+    Then I should see "/img/medology.png" image in "dynamic-image"
+
+  Scenario: Developer Can Test if an Image without src loaded dynamically
+    When I reload the page
+    Then I should see "/img/medology.png" image in "dynamic2-image"
```
jose-ochoa: I think this should assert that an image is not loaded and src is empty, only upon scrolling into view is the image loaded and src has a value.

rashidsaf: What you describing is lazy loading and it is not a part of the change. It is not by default and we may choose or not choose to lazy load the images on STDCheck with using the plugin. So we cannot test that here. But I am simulating the src load by javascript delay.

jose-ochoa: The problem I see is that this test will still pass if in dynamic2-image the src attribute is populated. Maybe we don't need this test since it is simply adding the waitfor?

Issue #325, opened 2020-03-12
False positive assertPageAddress


PayneWheat:
## Describe the bug
False positive assertion on behat tests where a link is clicked and the resulting page is verified to be a specified URL. I uncovered this when I changed a link's href, and then found a behat test asserting the proper webpage has opened was still passing with the old URL. 

## To Reproduce
1. Find any usage of `a new tab or window should open at...` in features directory that specifies a URL (paths seem to be working as intended). 
2. Swap The URL specified with something like `https://hello`, which obviously isn't a valid URL nor the href of the link being clicked.  
3. Run behat test.

## Expected behavior
Test should fail.

## Actual behavior
Test passes. Looking closely at the code, I believe the issue lies in `FlexibleContext::assertPageAddress` (line 80 in src/Behat/FlexibleMink/Context/FlexibleContext.php). 
`if (!strpos($actual, $page) === 0) {`
Either Wrapping the expression in parentheses with the `!` operator before the opening paranthesis or changing the comparison operator to `!==` seems to work.

johnnie502:
### Local QA
Since this a change for the automating test framework, there's no way to test it manually. I have run the new behat test locally, it passed as expected

Move to Test & Deploy

fiyyazp:
will get with rodolfo today to deploy

fiyyazp:
will get a pr out today

fiyyazp:
will get a PR out today

fortoulrodolfo:
### Pre Production QA - https://github.com/Medology/stdcheck.com/issues/10013

- Ran behat test successfully. Updated URL path and re-ran it, the test fails as expected.
- Moving to **Deploy&Test**

Chekote:
Will deploy and test today.

BKorenek:
Only testing updated, moving to Approval https://github.com/Medology/stdcheck.com/issues/10021

PayneWheat:
Scored by @DevanTarunKumar and @manuerumx

Chekote:
In addition to fixing the bug, you've actually changed the way this method works in an inappropriate way. I've moved this back to in-progress and [commented](https://github.com/Medology/FlexibleMink/pull/330#discussion_r406188892) on the necessary changes.

juliohl:
### QA - Local
QA will not check FlexibleMink tickets as per instructions from @Chekote, need to confirm with him if this should be moved by us to Deploy & Test of someone else should QA this.

PayneWheat:
[...]

PR #329, opened 2020-03-23
Patching false positive page address assertion (ver1)

PayneWheat:
For #325 

## Describe the bug

## To Reproduce
3. Run behat test.

## Expected behavior
Test should fail.

## Actual behavior
Test passes.

## Solution

```
```

```
if ($actual !== $page) {
```

### Note:


PayneWheat:
Scored 8 by @DevanTarunKumar and @manuerumx

PR #330, opened 2020-03-23
Patching false positive page address assertion (ver2)

PayneWheat:
For #325 

## Describe the bug

## To Reproduce
3. Run behat test.

## Expected behavior
Test should fail.

## Actual behavior
Test passes.

## Solution

```
```

```
if ($actual !== $page) {
```

### Note:

### DIFF
--- a/src/Medology/Behat/Mink/FlexibleContext.php
+++ b/src/Medology/Behat/Mink/FlexibleContext.php
@@ -79,18 +79,33 @@ class FlexibleContext extends MinkContext
 
         /* @noinspection PhpUnhandledExceptionInspection */
         Spinner::waitFor(function () use ($page) {
-            // is the page a path, or a full URL?
-            if (preg_match('!^https?://!', $page) == 0) {
-                // it's just a path. delegate to parents implementation
-                parent::assertPageAddress($page);
-            } else {
-                // it's a full URL, compare manually
-                $actual = $this->getSession()->getCurrentUrl();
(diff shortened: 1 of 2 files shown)


PayneWheat:
Need to do composer update medology/flexible-mink to update composer.lock in repos that utilize Flexible Mink v2

src/Medology/Behat/Mink/FlexibleContext.php
```diff
@@ -79,20 +79,35 @@ public function assertPageAddress($page)
...
+    {
+        // is the page a path, or a full URL?
+        if (preg_match('!^https?://!', $page) == 0) {
+            // it's just a path. delegate to parents implementation
+            parent::assertPageAddress($page);
+        } else {
+            // it's a full URL, compare manually
+            $actual = rtrim($this->getSession()->getCurrentUrl(), '/');
+            $page = rtrim($page, '/');
+            if ($actual !== $page) {
```
Chekote, line 102: In addition to fixing the bug, you've actually changed the way this method works in an inappropriate way. This method is supposed to assert that the URL starts with the specified page. But you've modified it to check that the URL matches the specified page exactly. This will now force anyone who uses this method to specify all of the query parameters or hash string for any page that they want to assert.

Please change this to:

```php
if (strpos($actual, $page) !== 0) {
```

PayneWheat: New PRs are up

PR #331, opened 2020-03-23
Add a wait for when asserting the number of rows in HTML tables

aserv92:
For: https://github.com/Medology/stdcheck.com/issues/8707

This adds a spinner to the method that asserts the number rows in an HTML table.

### DIFF
--- a/src/Behat/FlexibleMink/Context/TableContext.php
+++ b/src/Behat/FlexibleMink/Context/TableContext.php
@@ -296,19 +296,21 @@ trait TableContext
             throw new InvalidArgumentException('Number of rows must be an integer greater than 0.');
         }
 
-        $table = $this->getTableFromName($name);
+        return $this->waitFor(function () use ($name, $num, $fullTable) {
+            $table = $this->getTableFromName($name);
 
-        $rowCount = count($table['body']);
+            $rowCount = count($table['body']);
 
-        if ($fullTable) {
-            $rowCount += count($table['head']) + count($table['body']);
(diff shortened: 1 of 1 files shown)


src/Behat/FlexibleMink/Context/TableContext.php
```diff
@@ -296,19 +296,21 @@ public function assertTableHasRows($name, $num = 1, $fullTable = false)
             throw new InvalidArgumentException('Number of rows must be an integer greater than 0.');
         }
 
-        $table = $this->getTableFromName($name);
+        return $this->waitFor(function () use ($name, $num, $fullTable) {
```
jose-ochoa: It is my understanding that waitFor would not be used on assertion methods on flexiblemink as you would force this wait on any other projects that implement this method. Please discuss with @Chekote

johnnie502: I don't think so 🤔 We are not suppose to use waitFor on assertions for very basic things like button exists, since other assertion may rely on it heavily. This assertion is more of a high level, not likely to be relied on.

Chekote: The topic is somewhat complex, but to put it simply, if spinners are nested it's not really a huge problem, because since spinners use a time-out instead of a max-attempts approach, any time consumed by a nested spinner will also consume the available time from the parent spinner, thus the parent will time-out just as quickly as if there was no nesting at all.

PR #333, opened 2020-03-25
Patching false positive page address assertion (ver1)

PayneWheat:
For #325 

## Describe the bug

## To Reproduce
3. Run behat test.

## Expected behavior
Test should fail.

## Actual behavior
Test passes.

## Solution

```
```

```
if ($actual !== $page) {
```

Some refactoring was necessary because `WebContext::assertNewWindowPageAddress` in STDcheck was calling `FlexibleContext::assertPageAddress` from within a `waitFor`, and `assertPageAddress` also uses a `waitFor`. So we were waiting on `assertPageAddress` to timeout before actually trying to change the tab/window.

I abstracted the logic from within the `waitFor` in `FlexibleContext::assertPageAddress` into `assertPageAddressOnce`, and I changed the callback in `WebContext::assertNewWindowPageAddress` to this new method. 

This was all to accommodate `callRecords.feature` in STDcheck, which seems to be the only test that was causing a false positive (Rodolfo said it was from a 3rd party extension CSA's no longer use). I have a PR for the changes that need to be applied to STDcheck.

### Note:


features/FlexibleContext/assertPageAddress.feature
```diff
@@ -32,3 +32,7 @@ Feature: Assert Page Address Method
       | Param      | Value        | Error Message                                                  |
       | invalidKey | test         | Query did not contain a invalidKey parameter                   |
       | key        | invalidTest  | Expected query parameter key to be invalidTest, but found test |
+
+  Scenario: External URL address
+    When I go to "https://www.google.com"
+    Then I should be on "https://www.google.com"
```
jose-ochoa: Please add a newline at the end of the file

PayneWheat: Sorry, I had missed this. Done

PayneWheat:
Will need to run `composer update medology/flexible-mink` on STDcheck and push changes to composer.lock on the PR for STDcheck

PR #339, opened 2020-04-09
Fixed URL assertion logic

PayneWheat:
For #325 
I created an issue by checking for strict equality when asserting a URL. We should only check if the actual page starts with the supplied URL.


juliohl:
### QA - Local
QA will not check FlexibleMink tickets as per instructions from @Chekote, need to confirm with him if this should be moved by us to Deploy & Test of someone else should QA this.

PR #340, opened 2020-04-14
Fix getListOfAllNodeElementParents includes stopAt node

Chekote:
For Medology/stdcheck.com#6717

This is a port of PR #257

# Problem:
The getListOfAllNodeElementParents method takes a stopAt argument, which was intended to prevent an ancestor matching the specified HTML tag from being included in the returned node elements. Instead, it was the last node element that was included in the returned list. The wording of the parameter name and the description did not make this clear. But the typical use case for this is to not include the body tag as it has some strange behavior regarding its viewport rectangle, overflow and whether or not elements outside of the rectangle are truly visible or not. Instead of fixing the body on pages that this problem manifests on, the developer decided to implement this approach and ignore the body instead.

# Implementation:
I ported the logic from PR #257 so that the method behaves this way in FlexibleMink 2.0 as well as 1.0.

### DIFF
--- a/src/Medology/Behat/Mink/FlexibleContext.php
+++ b/src/Medology/Behat/Mink/FlexibleContext.php
@@ -1769,7 +1769,10 @@ class FlexibleContext extends MinkContext
     public function assertNodeElementVisibleInViewport(NodeElement $element)
     {
         if (!$this->nodeIsVisibleInViewport($element)) {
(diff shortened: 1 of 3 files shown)


Chekote:
I just noticed the changes in this PR are a little messed up. I'm going to close this and re-open once they are fixed.

PR #341, opened 2020-04-14
Fix getListOfAllNodeElementParents includes stopAt node

Chekote:
For Medology/stdcheck.com#6717

This is a port of PR #257

## Problem:
The getListOfAllNodeElementParents takes a parameter called "stopAt". From the use cases in host projects, it seems that it was intended for the node that matches this parameter to not be included in the resulting node list. But the method was stopping at this node and included it, and only excluding all further parents. Flexible Mink 1.0 was already modified by PR 257 to modify the behavior so the stopAt node was not included in the returned list.

## Fix:
I implemented the exact same functionality as 257, but also took the opportunity to clean up the method and write unit tests for it.

### DIFF
--- a/tests/Medology/Behat/Mink/FlexibleContext/GetAncestorsTest.php
+++ b/tests/Medology/Behat/Mink/FlexibleContext/GetAncestorsTest.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Tests\Medology\Behat\Mink\FlexibleContext;
+
+use Behat\Mink\Element\NodeElement;
+use PHPUnit_Framework_MockObject_MockObject as MockObject;
+
+class GetAncestorsTest extends FlexibleContextTest
+{
+    private $body;
+    private $div;
+    private $list;
+    private $listItem;
+    private $button;
+
+    public function setUp() {
+        parent::setUp();
+
+        // Given I have an HTML body
+        $this->body = $this->mockNode('body');
+
+        // And the body has a div
+        $this->div = $this->mockNode('div', $this->body);
+
+        // And the div has an unordered list
+        $this->list = $this->mockNode('ul', $this->div);
+
+        // And the list has a list item
+        $this->listItem = $this->mockNode('li', $this->list);
+
+        // And the list item has a button
+        $this->button = $this->mockNode('button', $this->listItem);
+    }
+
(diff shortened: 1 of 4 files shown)


tests/Medology/Behat/Mink/FlexibleContext/GetAncestorsTest.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Creates a mocked NodeElement with an optional parent.
+     *
+     * @param  string                 $tagName the type of node element to mock
+     * @param  NodeElement|null       $parent  the optional parent for the node element
+     * @return MockObject|NodeElement
+     */
+    protected function mockNode($tagName, NodeElement $parent = null)
```
izazueta, line 68:
```suggestion
    protected function mockNode(string $tagName, NodeElement $parent = null)
```

tests/TestCase.php
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @param object|string $object     Instantiated object that we will run method on
+     * @param string        $methodName Method name to call
+     * @param array         $parameters array of parameters to pass into method
+     *
+     * @throws ReflectionException if there is a problem invoking the method
+     *
+     * @return mixed method return
+     */
+    public function invokeMethod($object, $methodName, array $parameters = [])
```
izazueta, line 22:
```suggestion
    public function invokeMethod($object, string $methodName, array $parameters = [])
```

tests/Medology/Behat/Mink/FlexibleContext/GetAncestorsTest.php
Chekote: Not possible. Flexible Mink 2.0 supports PHP 5.6. https://github.com/Medology/FlexibleMink/blob/31b089864c57e34bf17ad9759248f6619c2942cd/composer.json#L12

tests/TestCase.php
Chekote: Not possible. Flexible Mink 2.0 supports PHP 5.6. https://github.com/Medology/FlexibleMink/blob/31b089864c57e34bf17ad9759248f6619c2942cd/composer.json#L12

PR #347, opened 2020-05-20
Adding in pull approve config

BKorenek:
Adding in config extending the scripts repos base config. This configuration has been simplified to fix several issues we are having with pull aprove 

Note for QA: Pull requests are solely updating the config file for how we assign reviewers. Nothing in this PR will be testable.


BKorenek:
Closing while we determine if this is the solution we want.

PR #350, opened 2020-06-23
Fix bug with the type caster casting ints that are too big

aserv92:
**Method:**
Using a number as a string that is more than PHP_MAX_INT

**Expected Behavior:**
The number should have the same string value

**Actual Result:**
The number does not have the same string value if the number is more than PHP_MAX_INT

**Cause:**
The type caster's regex matches all numbers based on a regex.

**Solution:**
Add logic to the type caster to check if PHP_MAX_IN is less than the string value and return the string value if the int  would be bigger than PHP_MAX_INT


aserv92:
Switched reviewers to be synchronized with my other PR for the master branch
https://github.com/Medology/FlexibleMink/pull/351

PR #351, opened 2020-06-23
Fix bug with the type caster casting ints that are too big

aserv92:
**Method:**
Using a number as a string that is more than PHP_MAX_INT

**Expected Behavior:**
The number should have the same string value

**Actual Result:**
The number does not have the same string value if the number is more than PHP_MAX_INT

**Cause:**
The type caster's regex matches all numbers based on a regex.

**Solution:**
Add logic to the type caster to check if PHP_MAX_IN is less than the string value and return the string value if the int  would be bigger than PHP_MAX_INT

### DIFF
--- a/src/Medology/Behat/TypeCaster.php
+++ b/src/Medology/Behat/TypeCaster.php
@@ -18,7 +18,9 @@ trait TypeCaster
      */
     public function castStringToInt($string)
     {
-        return $string <= PHP_INT_MAX ? intval($string) : $string;
+        $intval = intval($string);
+
+        return strval($intval) === $string ? $intval : $string;
     }
 
     /**
--- a/src/Medology/Behat/TypeCaster.php
+++ b/src/Medology/Behat/TypeCaster.php
@@ -10,11 +10,14 @@ trait TypeCaster
     /**
      * Casts a step argument from a string to an int.
      *
-     * @Transform /^(0|[1-9]\d*)$/
+     * Will cast the string to an int if the string is int like and within the max int range of the system.
+     * Otherwise, the original string will be returned unmodified.
+     *
+     * @Transform /^((\-\d+)|\d+)$/
      *
      * @param string $string the string to cast
      *
-     * @return int|string the resulting int
+     * @return int|string the string cast to an int, or the original string if it is outside the max int range of the system
      */
     public function castStringToInt($string)
     {
--- a/src/Medology/Behat/TypeCaster.php
+++ b/src/Medology/Behat/TypeCaster.php
@@ -14,7 +14,7 @@ trait TypeCaster
      *
      * @param string $string the string to cast
      *
-     * @return int the resulting int
+     * @return int|string the resulting int
      */
     public function castStringToInt($string)
     {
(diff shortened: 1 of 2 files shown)


src/Medology/Behat/TypeCaster.php
```diff
@@ -14,11 +14,13 @@ trait TypeCaster
      *
      * @param string $string the string to cast
      *
-     * @return int the resulting int
+     * @return int|string the resulting int
      */
```
Chekote, line 18: Let's update the documentation to make the new behavior clear:

```
    /**
     * Casts a step argument from a string to an int.
     *
     * Will cast the string to an int if the string is int like and within the max int range of the system.
     * Otherwise, the original string will be returned unmodified.
     *
     * @Transform /^(0|[1-9]\d*)$/
     *
     * @param string $string the string to cast
     *
     * @return int|string the string cast to an int, or the original string if it is outside the max int range of the system
     */
```

tests/Medology/Behat/TypeCaster/CastStringToIntTest.php
```diff
@@ -0,0 +1,56 @@
...
+    public function conversionExamples()
+    {
+        return [
+            'numeric string values more than PHP_INT_MAX are not converted to integers'                   => ['9223372036854775808', '9223372036854775808'],
+            'numeric string values equal to PHP_INT_MAX are converted to integers'                        => ['9223372036854775807', 9223372036854775807],
+            'numeric string values less than PHP_INT_MAX are converted to integers'                       => ['9223372036854775806', 9223372036854775806],
+            'zero string values is converted to a integer'                                                => ['0', 0],
+            'low number string value is converted to a integer'                                           => ['1', 1],
+            'negative number string value is converted to a integer'                                      => ['-9999', -9999],
+            'negative numeric string values less than negative PHP_INT_MAX are not converted to integers' => ['-9223372036854775809', '-9223372036854775809'],
```
Chekote, line 53: You will need to update the `@transform` annotation of the method to support negative numbers. Without it, the method can be called programatically with a negative number argument, but Behat will never invoke the method to transform negative numbers from a feature file.

src/Medology/Behat/TypeCaster.php
```diff
@@ -10,11 +10,14 @@ trait TypeCaster
     /**
      * Casts a step argument from a string to an int.
      *
-     * @Transform /^(0|[1-9]\d*)$/
+     * Will cast the string to an int if the string is int like and within the max int range of the system.
+     * Otherwise, the original string will be returned unmodified.
+     *
+     * @Transform /^((\-\d+)|\d+)$/
```
Chekote, line 16: You have changed this so that it will now match a string that has a leading zero. This should not match that as it will break zip codes that have a leading zero by stripping it out.

You just want to add the optional minus sign before the non-zero repeating digit, because obviously negative zero is not a thing.

```suggestion
     * @Transform /^(0|-?[1-9]\d*)$/
```

Chekote: You had asked about being more productive before, and this was one of the things we went over. When making changes for code review feedback, you should only make the change that was requested. By making an additional change, you've caused yet more code review feedback, slowing down the process.

PR #356, opened 2021-02-23
Create formatter for phone numbers.

juanmrad:
### Story:

We don't have a way to format phone inputs on flexible mink store when the value displayed on front-end is different from the one capture on the DB when requesting.

### Solution:
Add a way to format phone numbers on US format and ###-###-#### as used on some places.
add tests.

### DIFF
--- a/src/Behat/FlexibleMink/Context/StoreContext.php
+++ b/src/Behat/FlexibleMink/Context/StoreContext.php
@@ -26,17 +26,14 @@ trait StoreContext
 
     protected static $dateFormat = DateTime::ISO8601;
 
-    protected static $FORMAT_MYSQL_DATE = 'a MySQL date';
-    protected static $FORMAT_MYSQL_DATE_AND_TIME = 'a MySQL date and time';
-    protected static $FORMAT_US_DATE = 'a US date';
-    protected static $FORMAT_US_DATE_AND_TIME = 'a US date and time';
-    protected static $FORMAT_US_DATE_AND_12HR_TIME = 'a US date and 12hr time';
     protected static $format_map = [
-        'a MySQL date'            => 'Y-m-d',
-        'a MySQL date and time'   => 'Y-m-d H:i:s',
-        'a US date'               => 'm/d/Y',
-        'a US date and time'      => 'm/d/Y H:i:s',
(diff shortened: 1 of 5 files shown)


src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -289,7 +286,7 @@ protected function parseProperty($property)
      *
      * @param  mixed  $property       the property to get from the object
      * @param  object $thing          the object to get the value from
-     * @param  string $propertyFormat the pattern for formatting the value.
+     * @param  mixed  $propertyFormat the pattern for formatting the value.
```
aserv92, line 289: `$propertyFormat` is nullable.

src/Behat/FlexibleMink/Context/StoreContext.php
```diff
@@ -314,7 +313,7 @@ protected function getValueForInjection($property, $thing, $propertyFormat = nul
      *
      * @param  string                   $propertyFormat the name of the property format to process.
      * @throws InvalidArgumentException if the property format is not supported.
-     * @return string                   the programmatic format.
+     * @return mixed                    the programmatic format.
```
aserv92, line 316: Isn't this also nullable?

tests/Behat/FlexibleMink/Context/StoreContextTest.php
```diff
@@ -496,6 +497,38 @@ public function testParseKey()
...
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider phoneNumberFormatDataProvider
+     *
+     * @param $input
+     * @param $output
+     */
```
aserv92, line 523: Missing type definitions and descriptions

aserv92 (changes requested):

src/Behat/FlexibleMink/Context/StoreContext.php
juanmrad: no. there is no way this will be null returned. as when it comes here propertyformat is set and will either return or throw exception.

aserv92: Yes, that looks correct. My mistake.

aserv92 (approved):

johnnie502 (approved):

PR #357, opened 2021-02-23
Create formatter for phone numbers.

juanmrad:
Story:
We don't have a way to format phone inputs on flexible mink store when the value displayed on front-end is different from the one capture on the DB when requesting.

Solution:
Add a way to format phone numbers on US format and ###-###-#### as used on some places.
add tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,79 +4,79 @@ commands:
   pull-merge:
     description: "Bring the repository up-to-date with master and handle setup"
     steps:
-    - checkout
-    - restore_cache:
-        key: composer-{{ checksum "composer.lock" }}
-    - run:
-        name: Configure the CircleCI git user
-        command: |
-          git config --global user.email "ci@medology.com"
-          git config --global user.name "Circle CI"
-    - run:
-        name: Configure git to use SSH instead of HTTP
-        command: |
-          git config --global url.git@github.com:.insteadOf git://github.com/
-          git config --global --add url.git@github.com:.insteadOf https://github.com/
-    - run:
-        name: Merge master
-        command: git fetch origin && git merge origin/master --no-edit
+      - checkout
+      - restore_cache:
+          key: composer-{{ checksum "composer.lock" }}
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -4,136 +4,142 @@ commands:
...
+          command: behat --strict -vv -f pretty -o std -f junit -o /tmp/artifacts
+      - store_test_results:
+          path: /tmp/artifacts
+      - store_artifacts:
+          path: /tmp/artifacts
+
+executors:
+  ubuntu:
+    machine:
+      image: ubuntu-1604:202007-01
```
aserv92, line 84: Consider using a newer image like `ubuntu-1604:202101-01`.
We should also consider overcoming the challenges related to the Ubuntu 20 images for CircleCI, as it seems there is only going to be one more Ubuntu 16 update.
https://circleci.com/docs/2.0/configuration-reference/#available-machine-images

src/Medology/Behat/StoreContext.php
```diff
@@ -266,7 +258,7 @@ protected function parseProperty($property)
      *
      * @param mixed        $property       the property to get from the object
      * @param array|object $thing          the thing to get the value from
-     * @param string       $propertyFormat the pattern for formatting the value
+     * @param mixed|null   $propertyFormat the pattern for formatting the value
      *
      * @return mixed the prepared value
```
aserv92, line 263: I know you haven't changed the return type of this method but it would be nice to document what the expected return types are. if possible.

aserv92 (changes requested):

johnnie502 (approved):

izazueta (approved):

src/Medology/Behat/StoreContext.php
juanmrad: it depends on the properties of the format mapper. `$format_map` right now it is just strings and arrays, but that could change.

aserv92 (approved):

PR #358, opened 2021-03-02
Add spinner to click link to retry when element becomes stale

jose-ochoa:
For: Medology/stdcheck.com#11256

BUG FIX:

**Cause:**
The click link step fails if the element becomes stale after it has been found in the DOM

**Solution:**
Add a spinner to re-retrieve the element and click on it.

Bonus: Upon running `init_project` discovered an error in Bash script about CI being an unbound variable and also CI was failing since submodules hadn't been updated in a while

### DIFF
--- a/bin/init_project
+++ b/bin/init_project
@@ -10,7 +10,7 @@ echo "Installing Composer..."
 echo "Installing Composer dependencies with out dev..."
 composer install --no-interaction --no-suggest
 
-if  [[ ! "$CI" == 'true' ]]; then
+if [[ "${CI:-}" == "true" ]]; then
     echo "Initializing git hooks..."
     "$ROOT"/bin/init_git_hooks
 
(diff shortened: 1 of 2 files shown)


aserv92 (changes requested):
Is updating the submodules required to resolve the issue?

juanmrad (changes requested):
The submodules did not fix the issue you describe on CI. that was caused by CI not importing the variable to the env from context. that is being fixed by another PR that I just merged. Your CI stopped failing because you fixed the style manually removing the need for the GITHUB OAUTH variable.

johnnie502 (commented):
If the problem has been resolved by other PR then we probably should close this one

jose-ochoa:
The submodules is required since the update fixed the following issue: https://app.circleci.com/pipelines/github/Medology/FlexibleMink/770/workflows/183f3022-ee12-47dc-83b7-2cc72933bd6e/jobs/2448/parallel-runs/0/steps/0-110?invite=true


The git hooks were not working. That was the second issue on CI.

juanmrad (approved):

johnnie502 (approved):

aserv92 (approved):

PR #390, opened 2021-09-13
Chaning time formats

BKorenek:
For: https://github.com/Medology/stdcheck.com/issues/13680

This will adjust time formats to show single digit hours instead of multiple. E.g. 06:12 PM vs 6:12PM


aserv92 (changes requested):
It seems like this change assumes that none of our applications have the desired functionality of showing times with double-digit hours

aserv92 (changes requested):
.

fortoulrodolfo (approved):

fortoulrodolfo:
@aserv92 Correct, but whenever we update the flexiblemink version in our projects, the tests are going to fail and we need to adjust that assumption there. (If we have test for everything :trollface: )

fortoulrodolfo (approved):

aserv92 (approved):

PR #400, opened 2021-09-29
Feature: Inject dynamic relative dates into steps

Chekote:
This feature exposes the functionality of `Carbon::parse` and allows developers to inject dynamic date/time values into their Behat steps.

e.g. if you wanted to assert the whenever you visit a page, the date of next Tuesday is somewhere on the page, you could do:

```cucumber
Then I should see (a date/time of 'Next Tuesday')
```

Note that transforms do not work on PyStringNodes or TableNodes, so if you want this feature to work with those, you are going to need to call `DateInjector::injectCarbonStrings()` on those manually.


Chekote:
Scrutinizer's complaints are all false positives complaining about dependencies missing when they are not. I have already filed a bug with them.

Chekote:
@BKorenek I'm just requesting a review from you first because I want to make sure you are happy with this implementation before I have anyone else take a look at it.

PayneWheat:
I'll hold off on reviewing this until Brandon takes a look.

src/Medology/Behat/DateInjector.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Returns a defensive copy of the format_map property.
+     *
+     * @return string[]
+     */
+    public function getFormatMap(): array {
+        return self::$format_map;
+    }
+}
```
BKorenek, line 66: WTB line break

features/DateInjector.feature
```diff
@@ -0,0 +1,22 @@
...
+    Given the value "The first Saturday was (a date/time of 'First Saturday')" is stored as "Message"
+    Then the "Message" should be "The first Saturday was 2022-10-29 00:00:00"
+
+  Scenario Outline: Date/Time can be injected in step in scenario outline
+    Given the value "<Message>" is stored as "Message"
+     Then the "Message" should be "<Expected>"
+
+  Examples:
+    | Message                                                          | Expected                                       |
+    | The first Saturday was (a date/time of 'First Saturday')         | The first Saturday was 2022-10-29 00:00:00     |
```
BKorenek, line 19: Will this work with the `formatted as ` functionality? that would be necessary on my end. Also, something like `now - 1 month`? The scenario I'm in is I select last month, and I need to assert that the beginning of last month appears on the page formatted as a MySQL timestamp and the end also

src/Medology/Behat/DateInjector.php
```diff
@@ -0,0 +1,66 @@
...
+    ];
+
+    /**
+     * Searches a string for date/time placeholders and uses carbon to replace them.
+     *
+     * @Transform /^(.*\(a date\/time of '(.+)'( formatted as '(.+)')?\).*)$/
+     * @param  string $subject the string to manipulate.
+     * @return string the subject with the date/time placeholders replaced.
+     */
+    public function injectCarbonStrings(string $subject): string {
```
BKorenek, line 34: now that I understand what the meaning of inject is, it's not so hard to wrap my head around but if we wanted to go with an alternative because I didn't get it we could do something like format string, parseStringTemplate etc. One of the biggest confusions I had tho I think was because it was a method on the store itself, so I conflated injection with injection into the store

src/Medology/Behat/StoreContext.php
```diff
@@ -32,22 +39,20 @@ class StoreContext extends Store implements Context
...
-        'a ###-###-### phone number'                    => ['/(\d{3})(\d{3})(\d{4})/', '$1-$2-$3'],
+        'a US phone number'          => ['/(\d{3})(\d{3})(\d{4})/', '($1) $2-$3'],
+        'a ###-###-### phone number' => ['/(\d{3})(\d{3})(\d{4})/', '$1-$2-$3'],
     ];
 
+    /**
+     * {@inheritdoc}
+     */
+    public function gatherContexts(BeforeScenarioScope $scope)
+    {
```
aserv92, line 50:
```suggestion
    public function gatherContexts(BeforeScenarioScope $scope): void {
```

src/Medology/Behat/Traits/GathersContexts.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * Gets the environment in order to do gathersContexts.
+     *
+     * This should be called at the beginning of any implementation of gathersContexts, and will return the environment.
+     *
+     * @throws RuntimeException if the current environment is not initialized
+     */
+    public function getEnvironment(BeforeScenarioScope $scope): InitializedContextEnvironment
+    {
```
aserv92, line 29: There are a few instances of the opening curly bracket being on the wrong line. Is the style fixer doing this? Maybe the style fixer isn't working on this project.

```suggestion
    public function getEnvironment(BeforeScenarioScope $scope): InitializedContextEnvironment {
```

tests/Medology/Behat/DateInjector/DateInjectorTest.php
```diff
@@ -0,0 +1,33 @@
...
+    public function setUp(array $methods = []): void
+    {
+        $this->dateInjector = $this->createPartialMock(DateInjector::class, $methods);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tearDown()
+    {
```
aserv92, line 29:
```suggestion
    public function tearDown(): void {
```

aserv92 (changes requested):
Just some minor style changes.

src/Medology/Behat/DateInjector.php
Chekote: The inject method on the store doesn't inject anything into the store. It injects values from the store into a string, just like this injects dates into a string. So it is consistent.

features/DateInjector.feature
Chekote: It should work with formatted as. I will add a test to cover that and ensure that it does.

I do not know if "now - 1 month" will work. As per our initial discussion of the acceptance criteria, I am simply exposing the functionality of the Carbon::parse method. So if that string works with Carbon::parse, then it will work here.

Chekote: Good catch on the `formatted as`. It wasn't working properly I'll have an update pushed in just a moment.

Chekote:
I have fixed all legitimate Scrutinizer complaints. All the remaining complaints are due to a Scrutinizer bug where it cannot find dependencies, which has already been reported to them.

aserv92 (approved):

aserv92 (commented):
Since you updated the Circle CI configuration and removed support for php5 can you please update the GitHub merge requirements for this repo after you merge this PR.
Screen Shot 2021-10-04 at 3 52 08 PM

PayneWheat (approved):

BKorenek (approved):

Chekote:
@aserv92 yeah, I was going to do that. I just didn't want to do so until this was ready to merge just in case anyone objected to dropping PHP5 support.

VitorTS (approved):

Issue #414, opened 2021-12-08
Ability to use stored data asserting CSV files content


oliveiraweb:
# Story

## Short Summary
`Short summary of the technical debt that makes this issue necessary`

E.g. We have a series of events that are no longer in use

## Acceptance Criteria
`A list of stasks required to consider this issue complete`

- [ ] Task 1
- [ ] etc

## Rationale: Pros vs Cons
`What is the rationale for the change? What are the pros and cons of the acceptance criteria`

### Pros
* Pro 1
* etc

### Cons
* Con 1
* etc

## Subject Matter Experts
`List of developers that would be considered to have in-depth info in or around the issue`

## Notes
`A list of resources or caveats that would be applicable to the issue`

## Exploratory Testing
`How is the end-user exprience affected by this (if at all)? Any special setup or tools required?`

## Announcement Message
`A message describing what was changed, why the change was made followed by tagging the author that created the issue`

HLCharlie:
working with Donald on CSFixer problem.

HLCharlie:
Sent to Andrew for Review; have one change and will send back to Review

PR #415, opened 2021-12-08
Update expected values on CSVContext to use stored values

oliveiraweb:
For: #414 

BUG FIX:
Now Developers can assert the CSV response contains values from the model.

**Solution:**
Implemented updateExpectedDataAndInjectValues function on CsvContext to update the expected value.


Chekote (changes requested):

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,22 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
             throw new Exception("CSV '$csvName' contains extra headers '" . implode("', '", $diff) . "'");
         }
     }
+
+    /**
+     * Update expected data injecting model values.
+     *
+     * @throws ReflectionException
```
aserv92, line 126: Please add missing information to this docblock.

```suggestion
     *
     * @param TableNode                 $table A table node containing the values to inject.
     * @throws ReflectionException When {the reason why the exception was thrown}
```

aserv92 (changes requested):
Nice job, just a minor change.

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,24 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
             throw new Exception("CSV '$csvName' contains extra headers '" . implode("', '", $diff) . "'");
         }
     }
+
+    /**
+     * Update expected data injecting model values.
+     *
+     * @param TableNode $table a table node containing the values to inject
+     *
+     * @throws ReflectionException When {the reason why the exception was thrown}
```
aserv92, line 128: Sorry if I confused you. The part that I put in curly brackets was meant to be changed to explain why the exception was thrown.

aserv92 (changes requested):

src/Medology/Behat/CsvContext.php
oliveiraweb: I forgot it haha, sorry

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,24 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
...
+     */
+    private function updateExpectedDataAndInjectValues(TableNode $table): array
+    {
+        $expectedData = array_map(function ($row) {
+            return array_map(function ($value) {
+                return $this->storeContext->injectStoredValues($value);
+            }, $row);
+        }, $table->getRows());
+
+        $tableNode = new TableNode($expectedData);
```
johnnie502, line 138: I am wondering if creating the TableNode object is necessary? It is converted to array when returns seems we don't need to let it go thru TableNode again

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,28 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
...
+     */
+    private function updateExpectedDataAndInjectValues(TableNode $table): array
+    {
+        $expectedData = array_map(function ($row) {
+            return array_map(function ($value) {
+                return $this->storeContext->injectStoredValues($value);
+            }, $row);
+        }, $table->getRows());
+
+        $expectedData = array_map(function ($row) {
```
johnnie502, line 138: Seems the code block has been duplicated, please remove line 138-143

johnnie502 (changes requested):

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,28 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
...
+            return array_map(function ($value) {
+                return $this->storeContext->injectStoredValues($value);
+            }, $row);
+        }, $table->getRows());
+
+        return array_values($expectedData);
```
Chekote, line 144: You could avoid the overhead of assigning and garbage-collecting a variable by using array_values inline:

```suggestion
        return array_values(array_map(function ($row) {
            return array_map(function ($value) {
                return $this->storeContext->injectStoredValues($value);
            }, $row);
        }, $table->getRows()));
```

If you are mapping an array using a function on an object, you can further simplify this and avoid the additional nested function by using an [object method callback](https://www.php.net/manual/en/language.types.callable.php):

```php
        return array_values(array_map(function ($row) {
            return array_map([$this->storeContext, 'injectStoredValues'], $row);
        }, $table->getRows()));
```

Unfortunately, the project currently supports < PHP 7.4, so we can't use [arrow functions](https://www.php.net/manual/en/functions.arrow.php) to simplify it further 😞

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,28 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
...
     }
+
+    /**
+     * Update expected data injecting model values.
+     *
+     * @param TableNode $table a table node containing the values to inject
+     *
+     * @throws ReflectionException When injected parameters are invalid
+     */
+    private function updateExpectedDataAndInjectValues(TableNode $table): array
```
Chekote, line 130: I don't feel like this is a good name for this method. Firstly, it seems to only be injecting stored values. I am not sure what you mean by "update expected data". Secondly, having the word "And" in a method is a [code smell](https://en.wikipedia.org/wiki/Code_smell) and indicates that you are violating [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) (which as far as I can tell, you aren't).

Also, I don't think this is a good location for the method, given what it is doing. It is injecting stored values into a TableNode. That seems like quite a useful and general-purpose piece of functionality, and therefore should not be hidden away inside something as specific as a CSV Context class.

Since the `injectStoredValues` is a method on the `StoreContext` class, and the `StoreContext` class is already something that is tightly coupled to Behat, I feel we should move this method to the `StoreContext` and rename it to `injectStoredValuesTable`

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,28 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
             throw new Exception("CSV '$csvName' contains extra headers '" . implode("', '", $diff) . "'");
         }
     }
+
+    /**
+     * Update expected data injecting model values.
+     *
+     * @param TableNode $table a table node containing the values to inject
+     *
+     * @throws ReflectionException When injected parameters are invalid
```
Chekote, line 128: There are many more exceptions that can bubble up from `injectStoredValues`. If you are not swallowing or transforming them in any way, you should have the exact same `@throws` here as the method(s) you are calling.

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,28 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
             throw new Exception("CSV '$csvName' contains extra headers '" . implode("', '", $diff) . "'");
         }
     }
+
+    /**
+     * Update expected data injecting model values.
```
Chekote, line 124: When you move this method, please also update this description:

```suggestion
     * Parses each TableNode cell for references to stored items and replaces them with the value from the store.
```

src/Medology/Behat/CsvContext.php
```diff
@@ -118,4 +119,28 @@ private function assertNoAdditionalHeaders(array $required, array $actual, strin
...
+     *
+     * @param TableNode $table a table node containing the values to inject
```
Chekote, line 126: The table does not contain values to inject. Values are being injected into it.

```suggestion
     * @param TableNode $table a table node containing the values to parse
```

Chekote (changes requested):

src/Medology/Behat/StoreContext.php
```diff
@@ -249,12 +250,34 @@ public function assertThingPropertyContains(string $thing, string $property, str
...
+     *                                  should not be possible.
+     * @throws ReflectionException      if $hasValue was provided but the reflection API says it does not exist. This
+     *                                  should not be possible.     */
+    public function injectStoredValuesTable(TableNode $table): array
+    {
+        return array_values(array_map(function ($row) {
+            return array_map(function ($value) {
+                return $this->injectStoredValues($value);
+            }, $row);
+        }, $table->getRows()));
```
Chekote, line 274: It seems you missed the second part of my [previous suggestion](https://github.com/Medology/FlexibleMink/pull/415#discussion_r767805228).

```suggestion
        return array_values(array_map(function ($row) {
            return array_map([$this->storeContext, 'injectStoredValues'], $row);
        }, $table->getRows()));
```

Chekote (changes requested):

Chekote (approved):

gamanuel (approved):

aserv92 (approved):

johnnie502 (approved):

aserv92 (approved):

johnnie502 (approved):

Chekote (approved):

gamanuel (approved):

PR #429, opened 2022-02-03
Bumping dependabot phpunit/phpunit from 5.7.27 to 9.5.10

Vagoasdf:
For: Medology/Umami#331

BUG FIX: Fixing the call of the  --strict-coverage option

**Cause:**
on the Unit tests of the CircleCi build, we were calling phpunit this way
`          command: phpunit --log-junit /tmp/artifacts/phpunit.junit.xml -strict-coverage`

**Solution:**
Adding an extra dash so it recognizes the options
`          command: phpunit --log-junit /tmp/artifacts/phpunit.junit.xml --strict-coverage`


PayneWheat (approved):
Not sure why scrutinizer is complaining. So I approve 👍

WellingtonBraga (approved):

aserv92 (approved):

PR #438, opened 2022-03-16
Extract WaitProxy class for re-use

Chekote:
# Problem:
We want to implement the wait proxy pattern on the projects that use this Composer package.
The current implementation for the Wait Proxy AsyncMink doesn't allow for easy use of the logic that is implemented there.

# Fix:
Extract an abstract WaitProxy class from AsyncMink, and have AsyncMink extend it.
Also, rename AsyncMink to FlexibleContextWaitProxy to make it clear what the class does.


Chekote:
Forcing review because Scrutinizer's complaints are invalid. The classes it claims are present, and the methods it claims does not exist, do. It just doesn't understand `@mixin`.

fortoulrodolfo (approved):

gamanuel (approved):

fortoulrodolfo (approved):

PR #440, opened 2022-03-17
Extract NounStore class from StoreContext.

lorenzodalaqua:
For: https://github.com/Medology/Umami/issues/424

We are planning on reusing functionality from Behat tests to expose an API to create preconditions for tests from other projects that rely on the API in STDCheck. Most Behat tests in STDCheck rely on functionality from `StoreContext` in FlexibleMink.

To separate logic to create these preconditions from Behat context classes I have extracted the `Medology\Behat\NounStore` class with the functionality from `StoreContext` that may be reused outside of a Behat test suite. 

`StoreContext` previously extended the `Chekote\NounStore\Store` class, now the new `NounStore` is the one that extends the store, and backwards compatibility for current usage of the `StoreContext` class was kept by using the `__call` magic method to delegate calls to the new `NounStore` class.


tests/Medology/Behat/StoreContextTest.php
```diff
@@ -28,10 +26,6 @@ class StoreContextTest extends TestCase
     public function setUp(): void
     {
         $this->storeContext = new StoreContext();
-
-        $property = new ReflectionProperty(StoreContext::class, 'dateInjector');
-        $property->setAccessible(true);
-        $property->setValue($this->storeContext, new DateInjector());
     }
```
Chekote, line 29: The majority of the methods being tested here are now on the NounStore class rather than StoreContext. Please create a NounStoreTest class and move the appropriate tests there, and test the methods on that class directly.

For the methods that should be able to be proxied to NounStore through StoreContext, please add tests to the StoreContextTest class that mock out the NounStore, and then attempt to call the method and verify that the method was called on the mock.

Chekote (changes requested):
Looks good. We just need to update the testing to cover the new NounStore class and the new proxy feature of the StoreContext class.

tests/Medology/Behat/StoreContextTest.php
lorenzodalaqua: Taking a better look at these test classes, it seems there is a lot of unused code and a few methods aren't even tested, will add tests for all methods and add the  'Do not review'  for now until you approve it.

Chekote: FYI: Adding tests for the methods you moved is reasonable. But if you didn't touch a method that is missing tests, you don't have to write tests for it. Yes, we want to get decent coverage, but filling out missing tests for code you didn't touch is outside of your scope of work.

lorenzodalaqua: Ok, just thought since I was changing things I could improve the codebase, but I'll just separate the tests for each class, remove the unused code and leave things a bit cleaner then.

Chekote (dismissed):

src/Medology/Behat/NounStore.php
```diff
@@ -0,0 +1,321 @@
+<?php
+
+namespace Medology\Behat;
```
fortoulrodolfo, line 3: Since this class will be used for diff purposes, should we still keep it in the `/Behat` directory? 🤔

fortoulrodolfo (changes requested):
All looks good to me. I only have 1 question. Great work btw. 👍

PayneWheat (dismissed):

src/Medology/Behat/NounStore.php
lorenzodalaqua: I don't see a reason, will take it out of that directory like the Spinner class in this package, will make it clear this is not related to Behat. Not sure if the Medology namespace is okay to use, but I suppose it does not have a NounStore class yet.

fortoulrodolfo (approved):

fortoulrodolfo (dismissed):

src/Medology/Behat/NounStore.php
Chekote: Good catch @fortoulrodolfo 👏

Chekote (approved):

PayneWheat (approved):

lorenzodalaqua:
Requesting force merge because Scrutinizer issues seems to be dependencies are misconfigured somehow, it cannot find some referenced classes from composer packages.

PR #444, opened 2022-03-28
Add setNounStoreInstance method to StoreContext.

lorenzodalaqua:
For: https://github.com/Medology/Umami/issues/413

This is a follow up PR to https://github.com/Medology/FlexibleMink/pull/440.

The functionality added is meant to enable using the logic in Behat contexts that sets up preconditions for tests in a testing API. This API will exposed endpoints that create the preconditions in the STDCheck application for tests in other  projects that use the STDCheck API (e.g. my-account).

The added method provides a way to inject a NounStore instance for the StoreContext, needed when the NounStore is also used by other classes that rely on dependency injection separate to Behat, such as the  Laravel IoC container.


WellingtonBraga (dismissed):

src/Medology/Behat/StoreContext.php
```diff
@@ -23,7 +23,10 @@ class StoreContext implements Context
     /** @var NounStore The noun store instance */
     protected $store;
 
-    public function __construct($store = null)
+    /**
+     * Constructor for the Store Context.
+     */
```
Chekote, line 28: This docblock is not necessary. It adds no additional info beyond what can already be determined by looking at the method signature.

src/Medology/Behat/StoreContext.php
```diff
@@ -42,6 +45,17 @@ public function __call($name, $arguments)
         return $this->store->$name(...$arguments);
     }
 
+    /**
+     * Sets the noun store instance used by the StoreContext. Used when interacting with a NounStore that
+     * is resolved by an IoC container external to Behat.
```
Chekote, line 50: A doc block should have a single succinct line as the heading. Then if necessary, further elaboration on line 3 and beyond.

Chekote (changes requested):

fortoulrodolfo (approved):

fortoulrodolfo (approved):

Chekote (approved):
Looks good. But as we discussed, if the new approach we discussed for STDcheck works we won't need this, so please go ahead and close this if not needed.

WellingtonBraga (approved):

